Issue #123, opened 2025-07-25
Make `sfm.reference.lightgbm` work with `utils.split_sequential`


mikkokotila:
## Background

Currently, all the other SFMs use `utils.split.split_sequential` but `sfm.reference.lightgbm` breaks if it is made to use it. Currently it uses utils.split_sequential` which is far from perfect solution. 

## TODO

- [ ] Study the difference between `utils.split.split_sequential` and `utils.split_sequential.split_sequential`
- [ ] Study what needs to change in `sfm.reference.lightgbm` to make it work with `utils.split.split_sequential`
- [ ] Port `sfm.reference.lightgbm` to use `utils.split.split_sequential`

PR #429, opened 2026-04-10
feat: add proto cohorts implementation

### LINKED ISSUE #361: RFC-1013: Proto-Cohorts Implementation
## Objective

After a parameter sweep, users have no structured way to combine multiple decoders from the same architecture into a single predictive unit. They must either pick one decoder and accept its individual variance, or manually stitch together outputs in ad-hoc scripts. The Cohort system gives users a principled, standardised way to select and aggregate decoders so that the combined output is more robust than any single decoder — critical in a systematic bitcoin trading context where risk management depends on signal quality.

## Summary

The Cohort sub-system does not train models, it refers to #347 for actually training the models.

A Cohort wraps one or more decoders of the same architecture into a single callable that behaves identically to a standalone decoder: data goes in, a binary prediction and associated probabilities come out. Users select decoders by a combination of `experiment_id` and `permutation_id` after a parameter sweep to select one or more decoders from the experiment.

The Cohort aggregates their outputs using probability-weighted voting when probabilities are available, and falls back to majority voting over binary predictions when probabilities are not available. The Cohort integrates downstream as a drop-in replacement for a single decoder downstream, for example at a strategy manager layer.

## Is This Really Necessary?

Individual decoders from a parameter sweep exhibit variance. In systematic trading, acting on a single decoder's signal means accepting that variance as unmitigated risk. The Cohort is the minimal abstraction needed to let users reduce that variance without leaving the existing `predict()` contract. It sits on the critical path between decoder training and strategy execution — without it, every user who wants ensemble behaviour builds their own, inconsistently, and outside the system's guarantees. The scope is deliberately narrow (selection + weighing only, single decoder architecture per cohort).

## Named Concepts

**Decoder**: Any ML model that exposes a predict callable. Takes input data, returns binary predictions (0 or 1) and optionally probabilities. The system includes 15 pre-built architectures; users may add their own.

**Prediction**: A binary integer output (0 or 1) from a decoder, representing the model's classification of a sample.

**Probability**: Specifically P(1) — the probability that the true class is 1 (positive). This is a float in [0, 1]. A value of 0.3 means the decoder believes there is a 30% chance the sample is positive, implying it would predict class 0. This definition must be enforced across all reference architectures and clearly documented for users bringing custom models.

**Cohort**: A group of one or more decoders from the same architecture, selected by `permutation_id`, wrapped behind a single predict() interface. The Cohort is the atomic unit consumed downstream (e.g. by a strategy manager).

**Permutation ID**: A unique identifier for a specific decoder parameterisation produced during a sweep. Used to select which decoders enter a Cohort.

**Probability-weighted voting (primary aggregation)**: Each decoder contributes its P(1) value. These are averaged across decoders. The Cohort predicts 1 if the mean P(1) > 0.5, else 0. Ties (mean P(1) = 0.5 exactly) collapse to 0.

**Majority voting (fallback aggregation)**: Used when the architecture does not produce probabilities. Each decoder contributes its binary prediction. These are averaged. Same threshold and tie rule apply.

## Outline

**What will be delivered:**

A Cohort class that wraps decoder predict callables and exposes its own predict() method with the following behaviour:

*Selection.* Users provide a list of `permutation_id` values. The Cohort is constructed from the corresponding decoders. All decoders in a Cohort must be from the same architecture. Validation of this constraint is in scope.

*Aggregation — primary path (probabilities available).* The Cohort collects the P(1) value from each decoder for every sample, computes the mean P(1) across decoders, and thresholds at 0.5 to produce the binary prediction. Exact ties collapse to 0.

*Aggregation — fallback path (probabilities not available).* The Cohort collects the binary prediction from each decoder, averages them, and applies the same threshold and tie rule. The fallback is selected once per Cohort based on whether the architecture produces probabilities, not per-sample.

*Single-decoder Cohort.* When a Cohort contains exactly one decoder, aggregation is skipped entirely. The decoder's outputs are passed through unchanged.

*Output contract.* The Cohort's predict() returns by default only the aggregated binary prediction. Probabilities and metadata are available through explicit options:

**Default output**: `predictions` — aggregated binary call, one integer (0 or 1) per sample. This is the only return value unless the caller asks for more.

**Optional: probabilities** (`return_probs=True`): the raw P(1) values from each decoder in the Cohort, one value per decoder per sample. Available only when the underlying decoder architecture produces probabilities. When the Cohort is in fallback mode, this option is unavailable and requesting it raises an error.

**Optional: metadata** (`return_meta=True`): a structured object contextual information. The exact contents of this object are not yet finalised; the option is reserved so that the main part of the interface remains stable when metadata fields are added. The most immediate example here is the meta-data to indicate "what `1` means" for the underlying decoder.

This design means the simplest consumer calls `cohort.predict(X)` and gets back exactly what `model.predict(X)` gives: a binary prediction array. The Cohort is a true drop-in replacement for a standard `predict` callable. With the exception that consumers who want richer information opt in explicitly.

**How this differs from the current state:**

Today, decoders can only be consumed individually, and even that depends on custom code until #347 is completed. There is no standard mechanism to group them, aggregate their outputs, or present a combined signal. Users who want ensemble behaviour must implement it outside the system.

## Out of Scope

- **Decoder selection methods beyond manual `permutation_id` input.** Automated selection (e.g., based on backtest performance) is a future concern.
- **Cross-architecture Cohorts.** Mixing decoders from different architectures is not a Cohort concern. It may happen at a higher layer (e.g. a strategy manager) but that is outside Limen.
- **Alternative aggregation methods.** Other weighting schemes (e.g., performance-weighted, Bayesian model averaging) are future extensions. This work delivers probability-weighted voting and majority-vote fallback only.
- **Training or retraining.** The Cohort is inference-only. See #347 .
- **Threshold configuration.** The 0.5 threshold and tie-to-0 rule are fixed in this version.
- **Metadata schema finalisation.** The `return_meta` option is reserved in the interface. The exact fields will be defined in a follow-up before implementation begins.
- **Downstream integration specifics.** Limen is a standalone module. How any particular consumer (strategy manager, dashboard, logging pipeline) integrates with it is that consumer's concern.

## User Experience

A user completes a parameter sweep and reviews results. They identify a set of promising parameterisations by `permutation_id`. They create a Cohort by supplying those IDs. From that point, the Cohort behaves exactly like a single decoder — they call predict(), they get back a binary prediction. If they want probabilities or diagnostic metadata, they opt in through arguments.

If they select a single decoder, the Cohort is a transparent wrapper — no aggregation logic, no behaviour change. If they select multiple decoders, the Cohort aggregates silently and the user sees a single, combined signal.

The optional probability output lets users inspect disagreement across decoders when they want to, without burdening the default path. The output of a Cohort where all decoders agree looks different from one where they split 3-to-1, and that information is available on demand through input arguments.

The fact that `experiment_id` is required as an input argument, since all experiments are single architecture, guarantees single architecture Cohorts only.

Because Limen is a standalone module, it imposes no requirements on what the consumer does with the output. The predict() contract is the full interface.

## New Data

**Per prediction call, the Cohort produces (depending on options):**

- `predictions` (always): array of integers (0 or 1), one per sample. Same shape and type as a single decoder's output.
- `probabilities` (on request): for each sample, the raw P(1) from each decoder in the Cohort. The length per sample equals the number of decoders in the Cohort. Only available when the architecture produces probabilities.
- `metadata` (on request): structured diagnostic information. Schema to be finalised. Reserved in the interface now so the signature is stable.

**Cohort metadata (available at construction):**

- List of `permutation_id` values in the Cohort.
- Count of decoders.
- Architecture identifier.
- Aggregation mode (probability-weighted or majority-vote fallback).

No new persistent storage is introduced. The Cohort is a runtime construct. Persistence of Cohort definitions (which IDs, which architecture) is outside Limen's scope.

## Support

**Failure modes and handling:**

- *Invalid permutation_id*: Fail at construction with a clear error listing available IDs. Do not partially construct.
- *Decoder raises during predict()*: Propagate the error. The Cohort does not mask decoder failures. If one decoder in the Cohort fails, the entire predict() call fails. Partial results are not returned.
- *Unexpected probability values*: If a decoder returns P(1) outside [0, 1], the Cohort raises a validation error rather than producing silently wrong aggregations.
- *Requesting probabilities in fallback mode*: Raise a clear error explaining that the architecture does not produce probabilities. Do not return zeros or placeholders.

**Diagnostics:**

When `return_probs` or `return_meta` are used, the output itself serves as the primary diagnostic tool. Consumers are responsible for their own logging and analysis of this data. Limen does not provide a diagnostic UI.

## Market Desirability

Ensemble methods are standard practice in quantitative trading. Providing a first-class Cohort abstraction as a standalone module means users do not need to build their own aggregation layer. This is particularly relevant for users running large parameter sweeps — the Cohort turns a sweep from "pick the best one" into "combine the best ones," which is a meaningfully different and more sophisticated workflow.

Because Limen is standalone with a standard predict() contract, it is usable by anyone — not just users of a specific strategy framework. This maximises the addressable audience. For users bringing custom models, the clean predict() contract and P(1) standardisation lower the integration barrier.

## Risks

**P(1) convention not universally followed.** Some of the 15 pre-built architectures, or user-supplied models, may return probability of the predicted class rather than P(1). If this is not caught, the aggregation produces wrong results silently. *Mitigation:* Audit all 15 architectures before launch. Document the P(1) requirement prominently in the custom model integration guide. Consider a validation check at construction time (e.g., flag cases where P(1) < 0.5 when the prediction is 1, which would indicate the wrong convention).

**Tie-to-0 bias.** Collapsing ties to 0 introduces a systematic bias toward the negative class. In a trading context, this is a conservative default (bias toward not-trading), which is intentional. But users should be aware of it. *Mitigation:* Document the tie rule explicitly. A subset of this same drawback is how sometimes the model has been trained so that 0 is "positive" and 1 is "negative" which is against the assumption of the direction of the systematic bias caused by tie-to-0. 

**Single decoder failure kills the Cohort.** No graceful degradation. *Mitigation:* This is the correct behaviour for v1 — partial results in a risk-sensitive context are more dangerous than a clear failure. Revisit if operational experience shows this is too brittle.

## Drawbacks

**Fixed aggregation logic.** Probability-weighted voting treats all decoders equally — a decoder that performed brilliantly in backtesting gets the same weight as one that barely passed selection. This is simple and predictable, but leaves performance on the table. Adding performance-weighted aggregation later will require changes to the Cohort's internal logic, though the external predict() contract remains stable.

**Variable-length probability output.** When `return_probs=True`, the probability data changes shape depending on how many decoders are in the Cohort. Any consumer that processes this data (rather than just logging it) needs to handle variable-length structures. 

**Tie-to-0 is a policy decision baked into the aggregation layer.** If a different use case wants tie-to-1 (e.g., a more aggressive stance), it cannot be configured without modifying the Cohort. Acceptable for v1 but noted as a constraint to consider in the future.

**Two paths to a single-decoder sensor.** A user can produce a callable sensor for a single decoder either by using Trainer directly or by creating a Cohort with one permutation ID. Both produce an identical predict() callable with identical behaviour. This overlap is a deliberate tradeoff: Trainer is the natural path for users working with a single decoder (which is the common case), while Cohort exists for multi-decoder aggregation. Forcing all users through Cohort would add unnecessary conceptual overhead for the single-decoder case, and "cohort" semantically implies more than one. The cost is that users need to know both exist — but the shared predict() contract means choosing wrong has no functional consequence.

## Alternatives

**No Cohort abstraction — let users aggregate externally.** Lowest effort, but gives up standardisation and pushes ensemble complexity onto the user. Not acceptable for a trading system.

**Always return probabilities and metadata alongside predictions.** Simpler interface (no options to manage), but forces every consumer to handle data they may not need. The default-minimal approach respects the principle that Limen is a drop-in for any single decoder — the simplest call should look and behave identically.

**Return only the aggregated scalar probability (mean P(1)) instead of per-decoder values.** Simpler structure, fixed schema. But loses the disagreement signal, which is the primary diagnostic value of the probability output. The per-decoder values are low-cost to produce and high-value when needed.

**Weighted aggregation by backtest performance from day one.** More powerful, but requires the Cohort to know about backtest results, coupling it to the evaluation pipeline. Higher effort and higher risk of getting the weighting wrong. The equal-weight approach can ship now and performance-weighting can be layered on later without breaking the predict() contract.

## Unresolved Questions

- **P(1) audit status.** Have all already built and to be built reference architectures been confirmed to return P(1) rather than probability-of-predicted-class? Which ones return probabilities at all, and which will use the fallback path?
- **Cohort persistence.** Is there some piece here on versioning? Should the Cohorts sub-system also have its own version of manifest ->  or should it send up in the experiment manifest post-experiment since it's experiment-specific?
- **Metadata schema.** What fields belong in the `return_meta` output? 

## Success Metrics

**Primary — prediction contract integrity.** 100% of Cohort predict() calls return outputs that conform to the defined contract (correct types, correct shapes, P(1) in [0, 1], predictions in {0, 1}). Measured via automated integration tests and runtime validation.

**Primary — adoption.** Within 8 weeks of launch, >50% of active strategies use at least one multi-decoder Cohort rather than a single decoder. Measured via usage data.

**Secondary — signal quality.** Cohorts with 3+ decoders show lower prediction variance than single decoders on the same architecture, measured on held-out data. This validates the core value proposition that simple manual aggregation + simple weighting reduces noise.

## Business Release Date

TBA

## Design

@mikkokotila 

### Non-coding

- [ ] 

### Coding

- [ ]

weida0715:

## What does this PR change?

- Add `Cohort` constructor with strict experiment source resolution and permutation ID validation
- Enforce single-architecture selection across all requested `permutation_ids`
- Set Cohort aggregation mode at construction from architecture capability hints
- Add `Cohort.predict()` probability-weighted aggregation path with tie-to-0 behavior
- Add `Cohort.predict()` majority-vote fallback aggregation path with tie-to-0 behavior
- Add single-decoder passthrough short-circuit for direct member prediction parity
- Add `return_probs` option in `predict()` for per-member P(1) outputs
- Add `return_meta` option in `predict()` for structured inference metadata
- Add probability range validation on decoder `_probs` outputs
- Raise explicit error when `return_probs=True` is requested in fallback mode
- Add comprehensive Cohort tests for constructor validation, aggregation behavior, passthrough behavior, and output-option contracts
- Add integration test coverage validating Cohort as a decoder-compatible replacement at inference

Closes #361 

## Checklist


limen/cohort/README.md
```diff
@@ -1,48 +1,71 @@
...
-## Quick orientation
+## Package map
 
 ```text
 cohort/
-└── regime_pools.py   # Offline filtering, clustering, diversification,
-                      # online model loading, aggregation, and RDOP
+├── cohort.py                # Cohort constructor + aggregation logic
+├── regime_pools.py          # RDOP implementation
+├── cohort_usecase.py        # Realistic integration/smoke tests for Cohort behavior
```
zero-bang: The package map lists `cohort_usecase.py` and `cohort_walkthrough.ipynb` but neither file exists on this branch. Remove these two lines or add the files.

docs/Cohort.md
```diff
@@ -0,0 +1,205 @@
...
+
+```python
+from pathlib import Path
+
+from limen.cohort import Cohort
+from limen.experiment.experiment_core import UniversalExperimentLoop
+from limen.experiment.param_domain import ParamDomain
+from limen.experiment.trainer import Trainer
+from limen.sfd.foundational_sfd import logreg_binary as sfd_module
+from tests.stubs.stubs import StubStrategy
```
zero-bang: `StubStrategy` does not exist — only `StubPruningStrategy` is defined in `tests/stubs/stubs.py`. A user running this example gets an `ImportError`. The tests themselves correctly use `GridStrategy` from `limen.experiment.param_search`. Replace:

```python
from tests.stubs.stubs import StubStrategy
```

with:

```python
from limen.experiment.param_search import GridStrategy
```

and `StubStrategy(domain)` → `GridStrategy(domain)` on line 153.

zero-bang (changes requested):
Solid implementation. The core aggregation logic (probability-weighted and majority-vote with tie-to-0) is correct, constructor validation is thorough, tests are comprehensive (33 tests covering construction, aggregation, output contracts, edge cases, and integration), and the documentation is well-structured.

Two issues to fix before merge — both are quick.

zero-bang (approved):

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+            }
+
+        preds = self.predict(data)
+        return {'_preds': preds}
+
+    def predict(self,
+                X: Any,
+                *,
+                return_probs: bool = False,
+                return_meta: bool = False) -> np.ndarray | tuple:
```
mikkokotila: `Sensor.predict(data_dict)` is the established downstream inference surface, but `Cohort.predict()` introduces a new ndarray/tuple API here and moves the decoder-style dict contract into `__call__()`. That means downstream consumers can no longer swap a `Sensor` for a `Cohort` transparently, which conflicts with both RFC #361 and the stated intent that the cohort should make many sensors look like one.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+        '''
+        Run cohort members on input data and return aggregated binary predictions.
+
+        Args:
+            X (Any): Input batch to score with all cohort members. Accepts either
```
mikkokotila: The docstring says callers can pass raw test features, but the implementation below always wraps non-dict input as `{'x_test': X}`. That only works for architectures whose `predict()` needs `x_test` alone; for example `TabPFNBinary` requires `x_val`, `y_val`, and `x_test`. As written, `cohort.predict(X)` is not a valid contract for every architecture this PR documents as cohort-compatible.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+            raise ValueError(
+                'Probabilities are unavailable for this Cohort architecture.')
+
+        if not self._members:
+            raise RuntimeError(
+                'Cohort has no bound decoder members. '
+                'Use set_members(...) before predict().'
+            )
+
+        member_input = X if isinstance(X, dict) else {'x_test': X}
```
mikkokotila: This constructs one shared `member_input` and reuses it across every bound member. In Limen, different `permutation_id`s can change active features, derived columns, and scaler fit, and `Sensor` does not re-prepare input from its own `round_params`. A real multi-permutation cohort can therefore fail on schema mismatch or score some members on the wrong transformed space.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+
+        # Single-decoder cohort short-circuit: pass through unchanged predictions.
```
mikkokotila: The RFC defines single-member passthrough in terms of preserving decoder behavior under the cohort's binary prediction contract. This branch returns raw member `_preds` unchanged, which is only safe if every supported fallback architecture already emits `{0,1}`. The documented fallback architecture `XGBoostRegressor` does not, so a one-member cohort can return continuous values instead of binary predictions.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+            member_preds: list[np.ndarray] = []
+
+            for member in self._members:
+                result = member.predict(member_input)
+                if not isinstance(result, dict) or '_preds' not in result:
+                    raise ValueError(
+                        'Decoder in majority_vote mode must return a dict with _preds.'
+                    )
+
+                preds = np.asarray(result['_preds'], dtype=float)
```
mikkokotila: RFC #361 defines fallback as majority voting over binary predictions. Here we cast `_preds` to float and feed them straight into the vote, which changes the decision rule whenever a fallback architecture emits continuous scores rather than binary votes. For a real regressor member, this is score averaging, not majority voting.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+        )
+        self.metadata = metadata
+        self._members: list[Any] = []
+
+    def set_members(self, members: list[Any]) -> None:
+
+        if not members:
+            raise ValueError('Cohort members must be a non-empty list.')
+
+        self._members = list(members)
```
mikkokotila: After constructor-time validation of `permutation_ids` and architecture, `set_members()` accepts any non-empty list without checking permutation ids, experiment provenance, architecture, or even member count. That makes `permutation_ids`, `architecture_id`, `aggregation_mode`, and the returned metadata advisory rather than enforced.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,448 @@
...
+                               *,
+                               return_probs: bool,
+                               return_meta: bool) -> np.ndarray | tuple:
+
+        if not return_probs and not return_meta:
+            return predictions
+
+        out: list[Any] = [predictions]
+        if return_probs:
+            out.append(probs)
```
mikkokotila: The RFC describes probabilities per sample, with one decoder probability per sample entry. This returns the transposed shape instead: a list of per-member arrays. Downstream consumers written to the RFC contract would need to transpose the result before they can use it.

docs/Cohort.md
```diff
@@ -0,0 +1,205 @@
...
+- applies strict threshold `> 0.5`
+- exact ties resolve to class `0`
+
+## Output Contracts
+
+### `predict(X, ...)`
+
+Spec-pure contract:
+
+- `predict(X)` → `np.ndarray`
```
mikkokotila: This docs section teaches `predict()` as an ndarray/tuple API and `__call__()` as the decoder-compatible adapter. If the cohort is meant to make many sensors look like one sensor downstream, that is backwards: `predict()` itself needs to preserve the sensor-style payload, with any cohort-specific metadata being optional and additive.

docs/Cohort.md
```diff
@@ -0,0 +1,205 @@
...
+)
+uel.run(experiment_name='cohort_docs_example', n_permutations=2)
+
+# 2) reconstruct trained members from selected permutations
+trainer = Trainer(experiment_dir)
+permutation_ids = [0, 1]
+members = trainer.train(permutation_ids)
+
+# 3) prepare inference input from one selected round schema
+data_dict = trainer._manifest.prepare_data(trainer._data, members[0].round_params)
```
mikkokotila: This example prepares inference input from `members[0].round_params` and then reuses that one schema for the whole cohort. That only works if all selected permutations share identical feature/scaler schema, which the constructor does not guarantee. As written, the docs normalize the exact multi-member failure mode the RFC is meant to support.

tests/test_proto_cohort.py
```diff
@@ -0,0 +1,582 @@
...
+
+
+def test_probability_weighted_predict_aggregates_mean_p1():
+
+    with TemporaryDirectory() as tmpdir:
+        exp_dir = Path(tmpdir) / 'exp'
+        _run_real_experiment(exp_dir, n_permutations=1)
+
+        sensors, x_test = _train_real_members_and_input(exp_dir, [0])
+        sensors = [sensors[0], sensors[0]]
```
mikkokotila: Most of the multi-member tests build a 'cohort' by duplicating the same trained sensor object. That verifies aggregation math on repeated outputs, but it never exercises the real cohort case where different selected permutations can have different feature/scaler schema and different inference requirements.

tests/test_proto_cohort.py
```diff
@@ -0,0 +1,582 @@
...
+            assert 'no bound decoder members' in str(e)
+
+
+def test_probability_weighted_predict_rejects_majority_vote_mode():
+
+    with TemporaryDirectory() as tmpdir:
+        exp_dir = Path(tmpdir) / 'exp'
+        _run_real_experiment(exp_dir, n_permutations=1)
+
+        _patch_round_architecture(exp_dir, {0: 'xgboost_regressor'})
```
mikkokotila: This test rewrites the architecture id to `xgboost_regressor` but still binds logreg-trained sensors, so it never exercises the actual behavior of a fallback member. That allows the fallback path to pass even if majority-vote aggregation is wrong for real fallback architectures.

tests/test_proto_cohort.py
```diff
@@ -0,0 +1,582 @@
...
+
+        live_data = {'x_test': x_test}
+
+        sensor_result = base_sensor(live_data)
+        cohort_result = cohort(live_data)
+        cohort_pred = cohort.predict(live_data)
+
+        assert isinstance(sensor_result, dict)
+        assert isinstance(cohort_result, dict)
+        assert isinstance(cohort_pred, np.ndarray)
```
mikkokotila: This 'drop-in replacement' test currently locks in the opposite of the intended contract: it asserts that `cohort.predict()` returns an ndarray while only `cohort(...)` matches the sensor dict interface. If `Cohort.predict()` is supposed to replace `Sensor.predict()` transparently, this test enshrines the wrong behavior.

mikkokotila (changes requested):
A few parts of this PR still drift from RFC #361's intended contract that a `Cohort` should make multiple trained sensors look like one sensor downstream. The main issues are the predict-surface mismatch, invalid assumptions about shared inference input across permutations, and fallback behavior that is not actually binary voting for real fallback architectures. I've left the specific inline comments below.

mikkokotila:
I think also good to add a ticket for later work for adding a genuine regressor test. No need for this PR though.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,522 @@
...
+            result = member.predict(self._prepare_member_input(member, X))
+            if not isinstance(result, dict) or '_preds' not in result:
+                raise ValueError(
+                    'Single-decoder cohort member must return a dict with _preds.'
+                )
+
+            raw_preds = np.asarray(result['_preds'], dtype=float)
+            probs = None
+
+            if self.aggregation_mode == 'probability_weighted':
```
mikkokotila, line 243: [P1] Single-member probability cohorts are still not transparent wrappers. This path recomputes `_preds` from `_probs > 0.5` instead of passing the member result through unchanged. That breaks RFC #361's single-decoder passthrough rule and the downstream Sensor contract for threshold-tuned decoders like `TabPFNBinary`, where `_preds` need not equal `_probs > 0.5`; it also strips member-specific keys such as `optimal_threshold` and `val_score` from the payload.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,522 @@
...
+            if preds.shape != base_shape:
+                raise ValueError('Decoder outputs must share the same shape.')
+
+        preds_matrix = np.vstack(member_preds)
+        mean_vote = np.mean(preds_matrix, axis=0)
+
+        return (mean_vote > Cohort._VOTE_THRESHOLD).astype(np.int8)
+
+    @staticmethod
+    def _to_binary_votes(preds: np.ndarray) -> np.ndarray:
```
mikkokotila: [P1] Fallback voting still hard-codes the wrong binaryization rule for regressor members. `_to_binary_votes()` uses `preds > 0.5`, but `XGBoostRegressor.evaluate()` directionalizes with `preds > 0`. A cohort of real fallback members can therefore vote `0` on positive regressor outputs such as `0.1`, so the Cohort decision rule still disagrees with the architecture's own downstream direction logic.

limen/cohort/cohort.py
```diff
@@ -0,0 +1,522 @@
...
+                    raise ValueError(
+                        'Decoder in probability mode must return a dict with _probs.'
+                    )
+
+                probs = np.asarray(result['_probs'], dtype=float)
+                self._validate_probability_range(probs)
+                member_probs.append(probs)
+
+            preds = self._probability_weighted_vote(member_probs)
+            probs = np.mean(np.vstack(member_probs), axis=0)
```
mikkokotila: [P2] `return_probs=True` is still returning the aggregated mean P(1), not the raw per-decoder probabilities defined in RFC #361. The RFC contract is one P(1) per decoder per sample; averaging here removes the disagreement structure that `return_probs` is supposed to expose and locks downstream callers into a different shape.

tests/test_proto_cohort.py
```diff
@@ -0,0 +1,756 @@
...
+def test_validate_probability_range_rejects_non_finite_values():
+
+    try:
+        Cohort._validate_probability_range(np.array([0.2, np.nan, 0.8]))
+        assert False, 'Expected ValueError'
+    except ValueError as e:
+        assert 'finite values' in str(e)
+
+
+def test_cohort_is_drop_in_decoder_replacement_for_dict_input():
```
mikkokotila, line 710: [P2] Please add an explicit regression guard for the Trainer-produced `Sensor` <-> `Cohort` payload contract. The current drop-in test only covers logreg and checks `_preds`/`_probs`, so it will not catch single-member shape drift for architectures that return extra keys or use non-0.5 prediction thresholds. A good guard here would assert that a one-member cohort preserves the exact `Sensor.predict()` payload shape/keys for at least one non-trivial architecture.

mikkokotila (changes requested):
The branch is closer now, but a few contract-level gaps still remain between the current implementation, the RFC, and the intended Sensor drop-in behavior. I've left the specific inline comments below.

mikkokotila (approved):

PR #435, opened 2026-04-16
ci: publish releases to pypi

mikkokotila:
## What changed
- add a release-triggered GitHub Actions workflow that builds the package from the published release tag and publishes the distributions to PyPI using the existing `PYPI_USERNAME` and `PYPI_PASSWORD` secrets
- add the Limen docs URL to `[project.urls]` in `pyproject.toml` so PyPI can render a Documentation project link

## Why
Releases were being created without a matching package publication step, which meant shipping a GitHub release did not automatically update the PyPI package. This wires package publication directly to the release event so the published release and the published package stay in sync.

## Impact
Every published GitHub release will now build the package from its release tag and upload the resulting distributions to PyPI. The package metadata will also expose the docs site as a first-class project link.

## Validation
- built the package locally with `python -m build` in an isolated virtualenv
- confirmed the new workflow is scoped to `release.published`
- confirmed only `pyproject.toml` and the new workflow file are included in this change

## Notes
This PR intentionally leaves the existing `project.license` deprecation warning untouched because it predates this change and is unrelated to the PyPI release flow.

### DIFF
--- a/.github/workflows/pr_publish_pypi.yml
+++ b/.github/workflows/pr_publish_pypi.yml
@@ -1,0 +1,56 @@
+name: Publish Package to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build_distribution:
+    name: Build Distribution
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout released tag
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.release.tag_name }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install build tooling
+        run: python -m pip install --upgrade build
+
+      - name: Build package distributions
+        run: python -m build
+
+      - name: Upload distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+
+  publish_to_pypi:
+    name: Publish Distribution to PyPI
+    needs: build_distribution
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Download distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_publish_pypi.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: ${{ secrets.PYPI_USERNAME }}
+          password: ${{ secrets.PYPI_PASSWORD }}
```
zero-bang, line 56: **Non-blocking / follow-up:** PyPI now recommends [Trusted Publishing (OIDC)](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/) over username/password secrets. Switching would let you drop `PYPI_USERNAME`/`PYPI_PASSWORD` entirely — no secrets to rotate, and each publish is cryptographically tied to a specific workflow run. The change is small: add `id-token: write` to the job permissions, configure a Trusted Publisher on PyPI for this repo + workflow, and remove the `user`/`password` inputs from the action.

Not blocking this PR since the existing secrets are already in place and the workflow is correct as-is.

zero-bang (approved):
Workflow is correct, clean, and integrates well with the existing release pipeline. Approving.

One forward-looking note inline — worth considering Trusted Publishing as a follow-up to eliminate credential management overhead.

.github/workflows/pr_publish_pypi.yml
mikkokotila: Thanks — agreed that Trusted Publishing is the better end state.

I kept this PR on the existing `PYPI_USERNAME` / `PYPI_PASSWORD` path because those secrets are already configured and this workflow works immediately as-is. Switching to OIDC also requires configuring a Trusted Publisher on PyPI for this repo + workflow, and I didn't want to land a workflow change that could break release publishing until that external setup is in place.

Happy to do the OIDC conversion as a small follow-up once the PyPI-side configuration is ready.

PR #439, opened 2026-04-17
ci: Improve coverage and add a 92 percent coverage gate

mikkokotila:
## Summary
- add substantial behavior-first test coverage across features, indicators, data helpers, trainer internals, transforms, and regime-pool helpers
- wire the new tests into the custom `tests.run` harness so CI measures the same suite we run locally
- add a dedicated `PR Checks Coverage` job inside the existing PR validation workflow with a hard `92%` floor, without rerunning the full suite twice

## Verification
- `pytest -q tests/test_feature_library_context.py tests/test_regime_pools_helpers.py tests/test_indicator_and_data_helpers.py tests/test_historical_data.py tests/test_trainer.py`
- `LIMEN_COVERAGE_RUN=1 python -m coverage run -m tests.run`
- `python -m coverage report --include='limen/*'`
- `python -m ruff check`

## Notes
- no production code was changed; this PR is test/workflow only
- `limen/transforms/calibrate_classifier.py` remains intentionally untouched as the agreed holdout
- unrelated local `datasets/selector_catalog.json*` files were not included


.github/workflows/pr_checks_coverage.yaml
```diff
@@ -0,0 +1,52 @@
...
+          if [ -f requirements.txt ]; then
+            pip install -r requirements.txt
+          fi
+
+      - name: Enforce package coverage floor
+        env:
+          LIMEN_COVERAGE_RUN: '1'
+        run: |
+          python -m coverage run -m tests.run
+          python -m coverage report --fail-under=92
```
zero-bang: This entire workflow is a near-exact duplicate of `pr_checks_tests.yml` — same triggers, same SSH tunnel, same Python setup, same `coverage run -m tests.run`. The only addition is `--fail-under=92`. Every PR will run the full test suite twice, doubling CI time and SSH resource usage.

Fix: add `--fail-under=92` to the existing workflow's coverage report step (line 53 of `pr_checks_tests.yml`):
```yaml
      - name: Run limen tests with coverage
        env:
          LIMEN_COVERAGE_RUN: '1'
        run: |
          set -x
          python -m coverage run -m tests.run
          python -m coverage report --fail-under=92
          mkdir -p coverage-artifacts
          mv .coverage coverage-artifacts/coverage.db
```

Then delete this file.

zero-bang (changes requested):
Good batch of behavior-first tests — the coverage is meaningful and the assertions check real logic, not just that functions run without error. All imports verified against production modules.

One blocking concern: the new coverage gate workflow duplicates the existing test workflow and will run the full suite twice per PR. See inline comment for the fix.

Minor: all existing workflows use `.yml`; this one uses `.yaml`. Consider renaming for consistency.

**Forward-looking concern**: with a hard 92% floor, any new production module added without accompanying tests will block CI. That's presumably intentional, but worth calling out so it doesn't surprise the team on the next feature PR.

mikkokotila:
Addressed the workflow review feedback in the latest push:

- deleted the duplicate `pr_checks_coverage.yaml` workflow
- folded the 92% coverage gate into the existing `PR Validation` workflow as a separate `PR Checks Coverage` job, so the suite only runs once per PR
- explicitly install `pytest` in CI so the expanded test suite does not rely on runner image defaults
- kept `Publish Coverage Comment` in the PR flow and made it run even when the upstream test job fails, while only publishing when a coverage artifact is present

All three inline threads on the deleted workflow are now outdated/resolved. Re-requesting review once checks settle.

zero-bang (approved):

PR #442, opened 2026-04-18
feat: split calendar and cyclical time features

mikkokotila:
## Summary
- replace `time_features` with two first-class feature files: `calendar_time_features` and `cyclical_time_features`
- export the new feature callables from `limen.features` and remove the old `time_features` surface
- update feature docs and package docs to reflect the new time-feature split
- rename and strengthen the feature tests to cover both discrete calendar fields and cyclical encodings

## Testing
- `source .venv-codex/bin/activate && pytest -q tests/test_feature_library_regimes.py`
- `git diff --check`


zero-bang (approved):
LGTM — clean split of `time_features` into `calendar_time_features` and `cyclical_time_features`.

Verified:
- Cyclical encoding math is correct (periods, zero-basing, sin/cos pairs)
- `is_weekend` correctly uses Polars `dt.weekday() >= 6` (Sat=6, Sun=7)
- All references to the old `time_features` are updated (tests, run.py, docs, README)
- `time_features` was never in `__init__.py.__all__`, so no downstream surface removed — only new exports added
- Tests are thorough: two dates crossing the weekend boundary, identity assertions on re-exports, and exact cyclical value checks

PR #447, opened 2026-04-18
Fix snapshot backtest alignment and trade metrics

mikkokotila:
## Summary
- shift snapshot execution to the next tradable bar by default and make the new lag semantics explicit in experiment and inline backtests
- compute run-level `trade_*` metrics by default, add explicit `bar_*` metrics, confusion-bucket mean return fields, and `mean_kelly_pct`
- populate inline snapshot `tp/fp/tn/fn_mean_return_pct` from the already-aligned test labels, with regression SFDs passing explicit directional actuals and hard one-to-one alignment checks

## Scope Note
- This PR is intentionally limited to the snapshot metric surfaces: `backtest_snapshot`, inline `backtest_*`, and post-run experiment backtest results.
- It does **not** include any `experiment_core` / `results.csv` resume handling for all-NaN metric columns.
- That boundary is a maintainer decision for this PR: mainline experiment-core behavior stays untouched here, and any future resume-schema work should be proposed separately if it is ever worth doing.

## Risk Note For Reviewers
- This PR includes a high-risk alignment change in `ReferenceModel._compute_backtest`.
- Inline snapshot backtests now pass the already-aligned test labels straight through to `backtest_snapshot` without adding any extra shift at this layer.
- Binary SFDs pass `y_test` as-is; regression SFDs pass an explicit directional `actuals` series derived from `y_test > 0`.
- Please review the label and price alignment contract carefully, especially the one-to-one assumptions between `preds`, `y_test`, and `price_data_for_backtest`, because an incorrect extra shift here would silently leak or invalidate the new confusion-bucket return metrics.

## Testing
- `uv run --with pytest pytest tests/test_reference_architecture.py -q`
- `uv run --with pytest --with coverage coverage run -m tests.run`


CHANGELOG.md
```diff
@@ -663,3 +663,11 @@
 ## v1.53.0 on 13th of April, 2026
 
 - Switch Limen spot raw-trade and kline queries from `tdw.binance_trades` to `tdw.binance_trades_complete` so `HistoricalData` sees the daily overlay view while preserving finalized monthly history
+
+## v1.53.1 on 18th of April, 2026
```
zero-bang: This branch was forked before `main` advanced to v2.0.0. Main's CHANGELOG already has entries for v2.0.0, v2.1.0, v2.1.1, and v2.2.0 after the v1.53.0 entry. After rebasing, this should be **v2.3.0** (not v1.53.1), and `pyproject.toml` needs to be bumped to match.

The version bump should arguably be minor (not patch) regardless, since changing `trades_count_mode` default from `'bars'` to `'runs'` and adding `execution_lag_bars=1` are behavioral breaking changes for both internal callers (`_experiment_backtest_results` and `ReferenceModel._compute_backtest`), which use defaults.

zero-bang (changes requested):
The backtest logic itself is solid — I traced through the math for execution lag shifting, run-level trade aggregation, confusion-bucket returns, and Kelly estimation, and everything checks out. Tests are well-structured and cover the key behaviors.

However, this branch needs a rebase before it can merge. It was forked before `main` advanced through v2.0.0 → v2.2.0 (HistoricalData rewrite, time features, wrangle removal, OHLCV features). Concretely:

- Tests cannot run on this branch because `limen/log/log.py` still has `import wrangle`, which was removed in v2.1.1 on main.
- CHANGELOG.md will conflict — main already has entries for v2.0.0 through v2.2.0 after v1.53.0.
- `pyproject.toml` was not updated (required per repo rules). Main is at `version = "2.2.0"`, so this release should be v2.3.0 (given the default-changing nature of the backtest updates).

After rebasing, the code changes themselves look good to merge.

**Forward-looking concern**: Existing experiment results stored from `_experiment_backtest_results` (which calls `backtest_snapshot` with defaults) will silently differ from new runs due to the `trades_count_mode` and `execution_lag_bars` default changes. Worth considering whether stored results should carry an `execution_lag_bars` tag to make old-vs-new comparisons obvious.

mikkokotila (commented):
## Review — Fix snapshot backtest alignment and trade metrics

Meticulous read of the diff at `codex/backtest-snapshot-fixes` (`0cb74b5`) and a manual run of `tests/test_backtest_snapshot.py` against the updated `backtest_snapshot`. All 7 new tests pass locally (Python 3.13.2, pandas 3.0.2, numpy 2.4.4).

### Overall

The core intent — fix look-ahead bias, promote trade metrics to a run-level view, add confusion-bucket returns and Kelly — is right and valuable. The math is correct, validation is added, and docs are updated in the right places. What needs attention before merge is mostly about the *behavioral break* this introduces for downstream consumers, and about semver hygiene after the PR falls behind `main`.

### Strengths

- **Causality fix.** Old default (`execution_lag_bars=0` implicit) evaluated `r_entry = price_change / open` on the same bar as the prediction, which is anti-causal — the model can't have seen `close` when `open` prints. `execution_lag_bars=1` makes the first tradable bar the one after the decision and aligns `r_entry` and `r_cont` consistently. Verified by tracing the shifted `open_px / close_px / dpx` through `entry_mask`, `cont_mask` and `exit_mask`.
- **Run-level trade metrics.** `trade_returns = ((1 + R_net[pos]).groupby(run_ids[pos]).prod() - 1)` correctly compounds net per-bar returns inside each consecutive `1`-run and includes the exit-bar cost.
- **Confusion bucket returns.** TP/FP/TN/FN masks use `valid_actual = actual.isin([0,1]) & aligned_bar_return.notna()`, which gracefully skips non-binary actuals and NaN rows. `.replace([np.inf, -np.inf], np.nan)` is the right belt-and-braces against `open_px == 0` after the shift.
- **Kelly.** Standard formula `f* = p - q/b` with `b = avg_win/avg_loss`, returns NaN unless both wins and losses exist and `payout_ratio > 0`. Verified with the existing test input (`win=0.20, loss=0.10 → f* = 25%`).
- **Validation.** Upfront `ValueError` on bad `trades_count_mode` and negative `execution_lag_bars` — good.
- **Docs + CHANGELOG in sync.** `docs/Backtest.md`, `docs/Log.md`, `limen/backtest/README.md`, and the `_experiment_backtest_results` docstring all list the new columns in the same order as the output DataFrame.

### Must address

1. **Versioning collision after rebase.** The entry adds `v1.53.1 on 18th of April, 2026`, but `main` is already at `v2.2.0` (time features + OHLCV families merged since this branch diverged). Because `trade_*` semantics change (see #2), this is not a patch bump in spirit either. Suggest `v2.3.0` with an explicit "behavioral change" note — plus resolving the CHANGELOG conflict against current `main`.

2. **Silent semantic change of `trade_*` columns is a compatibility break.** Before this PR, `trade_win_rate_pct` / `trade_expectancy_pct` / `trade_return_mean_win_pct` / `trade_return_mean_loss_pct` were *per in-market bar*. After this PR (with the new default `trades_count_mode='runs'`) the same column names carry *per run* semantics. Any stored experiment-backtest results written under the old code are no longer comparable to new results under the same column name. `trades_count_mode='bars'` lets users opt back in, but that's not the default. Options:
    - Keep the behavioral change (preferred — it's the more correct view) and call it out loudly in the CHANGELOG, *and* bump minor/major accordingly.
    - Alternatively, add the run-level values under new column names (e.g. `run_win_rate_pct`) and keep `trade_*` as the per-bar fields for one release before flipping the default.

3. **SFD reference-architecture caller is not updated.** [limen/sfd/reference_architecture/base.py:135](limen/sfd/reference_architecture/base.py#L135) calls `backtest_snapshot(bt_input)` with no kwargs. The `backtest_*` metrics emitted by every SFD using this reference architecture therefore silently switch to `execution_lag_bars=1` + run-level semantics. [tests/test_inline_metrics.py:42](tests/test_inline_metrics.py#L42) and [tests/test_reference_architecture.py:71](tests/test_reference_architecture.py#L71) only assert key presence, so they will not catch this. Pick one and make it explicit:
    - If you want the new semantics everywhere, document it in the CHANGELOG as "applies to SFD `inline_metrics` too".
    - If you want to limit the blast radius for this PR, pin the caller to `backtest_snapshot(bt_input, execution_lag_bars=0, trades_count_mode='bars')` for now and migrate SFDs separately.

### Should address

4. **Empty DataFrame raises `IndexError` at `eq_net.iloc[-1]`.** Confirmed by running with `pd.DataFrame({'predictions': [], 'open': [], 'close': [], 'price_change': []})` — `IndexError: single positional indexer is out-of-bounds`. Not hit in the current call sites, but trivial to guard: short-circuit to a NaN-filled row when `len(df) == 0`.

5. **Predictions on the last `execution_lag_bars` rows are silently dropped.** Because `pos` masks on `open_px.notna()` after the shift. This is causally correct but worth one line in `docs/Backtest.md` — users eyeballing `trade_runs_count` after setting `execution_lag_bars=N` may otherwise be confused when the last `N` `1`s "vanish".

6. **Confusion-bucket returns depend on `actuals` labeling convention.** `aligned_bar_return = dpx/open` uses the *shifted* prices, so the measured return is that of the executing bar (row `t+lag`). If `actuals[t]` is a *forward* label ("did bar t+lag go up?"), alignment is correct. If someone fed in a *coincident* label (bar t's own direction), TP/FP/TN/FN returns silently refer to a different bar than the label describes. A one-line callout in `docs/Backtest.md` under the confusion-bucket paragraph would help.

7. **Kelly reported is full Kelly.** `mean_kelly_pct` is the full-Kelly fraction, not half-Kelly. That's fine as a diagnostic but worth explicitly calling out in the docstring and `docs/Backtest.md` so users don't size directly from it. Also, the pre-filter `kelly_returns[kelly_returns != 0]` drops zero-return trades from the sample; mention this briefly because it nudges both `win_rate` and `payout_ratio`.

8. **Rounding precision is inconsistent across output columns.**
    - `trade_expectancy_pct`, `bar_expectancy_pct`, `tp/fp/tn/fn_mean_return_pct`, `mean_kelly_pct` → 3 decimals
    - `trade_return_mean_win_pct`, `trade_return_mean_loss_pct`, `bar_return_mean_win_pct`, `bar_return_mean_loss_pct` → 1 decimal
    - `trade_win_rate_pct`, `bar_win_rate_pct`, `bars_in_market_pct`, `max_drawdown_pct`, `total_return_*_pct` → 1 decimal

    Either harmonize everything to 3 decimals, or at least make the "mean win/mean loss" pair match their respective expectancy precision.

### Test coverage gaps (nice-to-have)

- No test for `execution_lag_bars >= 2` (verified manually; works).
- No test for `fee_bps > 0` on the new paths — cost-on-exit still matters and is untested in the new suite.
- No test for `bars_in_market_count == 0` / `trade_runs_count == 0` (confirmed the NaN-safe path works: `kelly_returns = pd.Series(dtype=float)` short-circuits correctly thanks to Python's `if`-expression short-circuit in `trade_returns if trade_runs_count else ...`).
- No test that `trade_runs_count` excludes the last-bar `1` when `execution_lag_bars=1`.

### Nits

- `pytest.raises(..., match="trades_count_mode")` uses double quotes; house style is single quotes except in f-strings.
- `aligned_bar_return` is computed unconditionally (line 132) even when `actual is None`; pushing it inside the `if actual is not None:` block saves one Series allocation. Negligible.
- `round(bar_expectancy_pct, 3)` etc. on `np.nan` keeps NaN, fine — but `int(bars_in_market_count)` sharing a row with NaN columns silently upcasts the whole row to float when accessed via `.iloc[0]`. Also fine, just a pandas quirk worth knowing.

### Verification evidence

- Ran `pytest tests/test_backtest_snapshot.py -v` against the PR branch: 7/7 pass in 10.49s.
- Spot-checked with a long single run (`preds=[0,1,1,1,0]`, fee=5bps, slip=5bps, lag=0): `total_return_gross_pct=15.0`, `total_return_net_pct=14.8`, `cost_round_trip_bps=20`, `trade_expectancy_pct=14.78` — matches the closed-form product of `(1+R_net)` per bar.
- With `execution_lag_bars=2` on `preds=[1,0,0]` and `open/close/pc` crafted so row 2 is the executable bar, `total_return_net_pct=10.0`, `tp_mean_return_pct=10.0`, `bars_in_market_count=1` — consistent with "prediction at t executes at t+2".
- Confirmed that a trailing `1` prediction with `execution_lag_bars=1` is excluded from `pos` because the shifted prices become NaN on the last row — causally correct, worth a doc line.

### Recommendation

Approve with the must-address points resolved: bump the changelog version to a v2.x slot, call out the `trade_*` semantic change explicitly, and make an intentional choice for the SFD reference-architecture caller rather than letting it flip silently. The should-address items are polish, not blockers.

limen/backtest/backtest_snapshot.py
```diff
@@ -34,37 +41,61 @@ def backtest_snapshot(df: pd.DataFrame,
...
+    if execution_lag_bars > 0:
+        open_px = open_px.shift(-execution_lag_bars)
+        close_px = close_px.shift(-execution_lag_bars)
+        dpx = dpx.shift(-execution_lag_bars)
+
     pos = (pred == 1) & open_px.notna() & close_px.notna() & dpx.notna() & (open_px != 0)
 
     bars_total = len(df)
+    bars_in_market_count = int(pos.sum())
     bars_in_market_pct = float(pos.mean() * 100.0)
```
mikkokotila: `execution_lag_bars` shifts the price series forward, but the backtest keeps the original row count and fills the now-untradable tail with `pos=False` / zero returns. Those non-existent bars still feed `bars_total`, `bars_in_market_pct`, and later `sharpe_per_bar`, so every lagged run is diluted and short windows can be materially wrong. In the two-row next-bar case, for example, there is only one tradable bar, but Sharpe is computed over `[ret, 0.0]` and exposure is reported against two bars.

tests/test_backtest_snapshot.py
```diff
@@ -0,0 +1,153 @@
+import math
+
+import pandas as pd
+import pytest
+
+from limen.backtest.backtest_snapshot import backtest_snapshot
```
mikkokotila, line 6: This file adds the right direct coverage, but the repo's canonical validation path is `python -m tests.run` in `.github/workflows/pr_checks_tests.yml`, and `tests/run.py` does not import or invoke any of these new tests. That means the added snapshot assertions will not run in CI or under the PR template's recommended local command, so the new regression protection is currently paper-only.

limen/backtest/backtest_snapshot.py
```diff
@@ -81,19 +112,76 @@ def backtest_snapshot(df: pd.DataFrame,
...
+    if trades_count_mode == 'bars':
+        trade_win_rate_pct = bar_win_rate_pct
+        trade_expectancy_pct = bar_expectancy_pct
+        trade_return_mean_win_pct = bar_return_mean_win_pct
+        trade_return_mean_loss_pct = bar_return_mean_loss_pct
+        kelly_returns = bar_returns
+    else:
+        kelly_returns = trade_returns if trade_runs_count else pd.Series(dtype=float)
+
+    kelly_returns = kelly_returns[kelly_returns != 0]
```
mikkokotila: `kelly_returns = kelly_returns[kelly_returns != 0]` removes zero-return trades/bars before estimating win and loss rates. That changes the empirical probabilities whenever breakevens exist and can materially inflate Kelly. For example, returns `[+20%, -10%, 0%]` give 25% here, but only 16.7% if the breakeven trade stays in the distribution. If `mean_kelly_pct` is meant to summarize the active return distribution, zeros need to remain in the denominator.

mikkokotila (commented):
I took a careful pass through the implementation, docs, and tests. I found three issues worth addressing before merge and left them inline below.

zero-bang (approved):

zero-bang (approved):

limen/log/_experiment_backtest_results.py
```diff
@@ -18,19 +18,28 @@ def _experiment_backtest_results(self: Any, disable_progress_bar: bool = False)
...
     all_rows = []
 
     for i in tqdm.tqdm(range(len(self.round_params)), disable=disable_progress_bar):
 
-        result_df = backtest_snapshot(self.permutation_prediction_performance(i))
+        result_df = backtest_snapshot(
+            self.permutation_prediction_performance(i),
+            execution_lag_bars=1,
+            trades_count_mode='runs',
+        )
```
mikkokotila, line 68: This now explicitly opts the log/experiment path into the new next-bar/run-level snapshot semantics, but it still passes raw `permutation_prediction_performance()` output straight into `backtest_snapshot()`. For regression SFDs, that table carries continuous predictions and continuous `actuals`, while the inline reference-architecture path added in the latest commit now binarizes both sides by direction (`preds > 0`, `y_test > 0`). That means the same regression round can still produce different backtest columns depending on whether the user reads `evaluate(..., inline_metrics=True)` or `uel.experiment_backtest_results`, and the log path will still leave the confusion-bucket return columns as `NaN` for regression rounds.

tests/run.py
```diff
@@ -240,6 +240,16 @@
...
+from tests.test_backtest_snapshot import test_snapshot_defaults_to_next_bar_execution
+from tests.test_backtest_snapshot import test_snapshot_can_reproduce_legacy_same_row_execution
+from tests.test_backtest_snapshot import test_snapshot_trade_metrics_are_run_level_by_default
+from tests.test_backtest_snapshot import test_snapshot_bar_mode_preserves_legacy_trade_metrics
+from tests.test_backtest_snapshot import test_snapshot_confusion_bucket_mean_returns_cover_all_quadrants
+from tests.test_backtest_snapshot import test_snapshot_confusion_bucket_respects_actual_col
+from tests.test_backtest_snapshot import test_snapshot_mean_kelly_pct_uses_trade_runs_by_default
+from tests.test_backtest_snapshot import test_snapshot_mean_kelly_pct_keeps_breakevens_in_denominator
+from tests.test_backtest_snapshot import test_snapshot_handles_empty_input
+from tests.test_backtest_snapshot import test_snapshot_validates_mode_and_execution_lag
```
mikkokotila, line 260: This runner now includes the new `test_backtest_snapshot` cases, but the last-commit regression guards in `tests/test_reference_architecture.py` and `tests/test_experiment_core_msq.py` still are not wired in here. There’s a second problem too: two of the new `tests/test_reference_architecture.py` additions use the `monkeypatch` fixture, while this canonical runner invokes each test as `test()` directly below, so they would not execute under `python -m tests.run` even if imported. The fixes from the latest commit are therefore still not covered by the suite CI actually runs.

mikkokotila (commented):
Fresh pass focused on the latest commit. The earlier snapshot math issues look fixed, but I still found one behavioral inconsistency between the experiment/log backtest path and the inline reference-architecture path, plus one remaining gap in the canonical `tests.run` coverage.

mikkokotila (commented):
## Re-review — after `a6c3493`, `d428a8a`, `95fe4f7`

Re-read the three follow-up commits against the earlier review, and re-ran the tests on `95fe4f7`:
- `pytest tests/test_backtest_snapshot.py` → 10/10 pass (3 new tests added)
- `pytest tests/test_reference_architecture.py` → 10/10 pass (3 new tests added)
- `pytest tests/test_experiment_core_msq.py tests/test_inline_metrics.py` → 18/18 pass (schema-recast regression + inline-metrics column added)

### Response to prior review

Previous points are all addressed except the rounding/stylistic nits I flagged as non-blockers:

| Prior point | Status |
|-------------|--------|
| Changelog version collision with `v2.x` | ✅ Bumped to `v2.3.0` in both CHANGELOG and `pyproject.toml` |
| `trade_*` semantic break called out | ✅ [docs/Backtest.md](docs/Backtest.md) and [docs/Log.md](docs/Log.md) both carry the behavioral-change notice |
| SFD reference caller silently changing | ✅ [limen/sfd/reference_architecture/base.py:156-160](limen/sfd/reference_architecture/base.py#L156) now passes `execution_lag_bars=1`, `trades_count_mode='runs'` explicitly; [_experiment_backtest_results.py:35-39](limen/log/_experiment_backtest_results.py#L35) likewise |
| Empty `DataFrame` → `IndexError` | ✅ Guarded via `tradable_R_net.empty`; `test_snapshot_handles_empty_input` covers it |
| Trailing-rows dropped with lag | ✅ Doc line added in the assumption list |
| Confusion-bucket labeling caveat | ✅ Added "if labels are coincident with the feature bar, the columns will refer to a different bar than the label" |
| Full-Kelly callout | ✅ Docstring + `docs/Backtest.md` now say "full-Kelly fraction". Also, the zero-return filter was removed entirely — breakevens count in the denominator now, and `test_snapshot_mean_kelly_pct_keeps_breakevens_in_denominator` pins the new semantics (≈ 16.667% in the 1 win / 1 loss / 1 breakeven bar-mode sample). |
| `aligned_bar_return` inside `if actual is not None` | ✅ Moved inside |
| Double quotes in `pytest.raises(..., match=...)` | ✅ Single quotes |
| Tests wired into `tests/run.py` | ✅ All 10 snapshot tests registered |

### Focus on `95fe4f7` — "Wire aligned actuals into inline snapshot backtests"

The design choice is clean: the per-row test labels already exist in `data['y_test']` (binary) or `(y_test > 0).astype(int)` (regression), and the caller simply passes them through `_compute_backtest(..., actuals=...)`. Snapshot then populates TP/FP/TN/FN automatically because the kwarg threads straight into the `'actuals'` column.

**Good:**
- The length pre-check `if len(price_pd) != len(preds)` and the matching `if len(actuals) != len(preds)` both raise `ValueError` with distinct messages — directly exercised by `test_compute_backtest_requires_aligned_lengths` (asserts on `match='price_data_for_backtest'` and `match='actuals'` separately).
- Monkeypatch-based tests (`test_logreg_inline_backtest_passes_binary_actuals_to_snapshot`, `test_xgboost_inline_backtest_passes_directional_actuals_to_snapshot`) capture the `df` handed to `backtest_snapshot` and assert both that `actuals` is present and that the values match `y_test` (binary) or `(y_test > 0).astype(int)` (regression). Clean isolation — they don't depend on the snapshot internals.
- `_compute_backtest` signature `actuals: np.ndarray | pd.Series | None = None` — clean `Optional`, matches Python 3.10+ minimum in `pyproject.toml`.

**Issues found in this commit:**

1. **Silent corruption when `y_test` contains NaN.** `np.asarray(actuals).astype(int)` in [base.py:149](limen/sfd/reference_architecture/base.py#L149) casts NaN to 0 with a `RuntimeWarning` (easy to miss) and the downgraded value is then indistinguishable from a genuine `0` label inside snapshot's `actual.isin([0, 1])` filter. Reproduced locally with `actuals=[1, NaN, 0]` and `preds=[1, 1, 0]`: row 1 (true label unknown) ends up in the FP bucket with its aligned return counted. Not catastrophic in practice — a clean prep pipeline should not hand NaNs to `y_test` — but it's a silent failure mode that the length validator does not catch. Either:
    - keep the cast as float and let snapshot's `isin([0, 1])` drop NaN naturally (one-line fix), or
    - add an explicit `if np.isnan(np.asarray(actuals, dtype=float)).any(): raise ValueError(...)` ahead of the int cast so NaNs surface loudly.

2. **Asymmetric test coverage across the four reference models.** `logreg_binary` and `xgboost_regressor` each have a dedicated monkeypatch test that captures the `actuals` passed to snapshot. `random_binary` and `tabpfn_binary` use the identical call pattern but have no direct actuals-pass-through test. Since the codepath is identical and `random_binary` has no optional-dependency issue, adding one small test mirroring the logreg one would close the coverage gap cheaply. `tabpfn_binary` is fair to skip because of the optional dep.

3. **Minor: `test_xgboost_inline_backtest_passes_directional_actuals_to_snapshot` doesn't assert on the kwargs.** `logreg`'s test checks `captured['kwargs']['execution_lag_bars'] == 1` and `captured['kwargs']['trades_count_mode'] == 'runs'`; the xgboost test only checks the `'actuals'` column. Minor — the kwargs flow through the same `_compute_backtest` so the check is redundant, but harmonizing would make future drift easier to catch.

### Focus on `d428a8a` — "Fix snapshot review issues and resume schema"

Two substantive things here beyond the review-response polish:

1. **`bars_total` and `bars_in_market_pct` semantics changed.** `bars_total = int(tradable.sum())` (was `len(df)`), and `bars_in_market_pct` now divides by `bars_total` instead of by row count. This is the right definition if "total" means "tradable total," but it's a *third* silent semantic change on top of the already-called-out `trade_*` one. Worth one more line in the changelog so users who were tracking `bars_total` historically aren't surprised. Also — test `test_snapshot_defaults_to_next_bar_execution` already pins `bars_total == 1` and `bars_in_market_pct == 100.0` for a 2-row, lag=1 input, which correctly encodes the new meaning.

2. **Resume schema recast in `_flush_results`** ([experiment_core.py:691-732](limen/experiment/experiment_core.py#L691)) is the right fix for the real problem: when a prior run wrote an all-null numeric column (e.g., `backtest_tp_mean_return_pct` not yet populated), CSV persistence reloads it as `pl.String` with values like `'nan'`, and the next batch's proper `Float64` column would fail `vstack`. The helper `_is_null_like_string_column` (values ∈ `{'', 'nan', 'null', 'none'}`, case-insensitive, after `strip()`) covers the common sentinels and the test `test_flush_results_recasts_all_null_resumed_string_columns` pins the String → Float64 direction.
    - The reverse direction (incoming String, existing Float) is handled symmetrically in code but not tested; a one-liner would be nice since it's the exact mirror case.
    - Not handled: new columns present only on one side. If a fresh batch introduces a column absent from the existing log (or vice versa), `vstack` will still complain. Likely out of scope for this PR but worth a TODO.
    - Sentinel coverage: `'NA'`, `'N/A'`, `'-'` are not in the null-like set. Polars' CSV reader typically turns empty/`NA` cells into proper nulls rather than string sentinels, so this is likely fine — just flagging in case you ever ship with a different reader.

### Focus on `a6c3493` — "Bump version"

Trivial and correct — `v2.2.1` → `v2.3.0` in `pyproject.toml`, matching the CHANGELOG. Nothing to flag.

### Remaining nits (unchanged from last round)

- Rounding precision is still heterogeneous across `*_pct` columns (1 dp for `trade_win_rate_pct`, 3 dp for `trade_expectancy_pct` / `bar_expectancy_pct` / `mean_kelly_pct` / `tp/fp/tn/fn_mean_return_pct`). Not a blocker.
- No test for `execution_lag_bars >= 2` (verified manually in the earlier review that it works).
- No test asserting `fee_bps > 0` behavior on the new run-level path. The cost-on-exit logic is unchanged from prior, so this is just a coverage gap, not a risk.

### Recommendation

LGTM to merge once #1 above (NaN in `y_test` → silent class-0 cast) is decided either way. The two candidates are a one-line defensive change (`dtype=float` + rely on snapshot filter) or an explicit `ValueError`. Everything else is either resolved or polish.

limen/sfd/reference_architecture/base.py
```diff
@@ -132,7 +132,11 @@ def _compute_backtest(self, preds: np.ndarray, data: dict) -> dict:
             'price_change': (price_pd['close'] - price_pd['open']).values,
         })
 
-        bt_result = backtest_snapshot(bt_input)
+        bt_result = backtest_snapshot(
+            bt_input,
+            execution_lag_bars=1,
+            trades_count_mode='runs',
+        )
```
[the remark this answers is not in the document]
mikkokotila: Addressed in `ReferenceModel._compute_backtest`. Inline snapshot backtests now pass aligned `actuals` through explicitly, enforce one-to-one length checks, reject NaNs before casting, and the canonical `python -m tests.run` path now covers binary plus directional-regression actuals pass-through.

tests/test_backtest_snapshot.py
mikkokotila: Addressed in the canonical runner. `tests/run.py` now imports the snapshot cases directly, includes the follow-up regression guards, and wraps the fixture-based reference-architecture tests with explicit `pytest.MonkeyPatch()` adapters so they really execute under `python -m tests.run`.

tests/run.py
mikkokotila: Addressed in `tests/run.py`. The canonical runner now includes the `experiment_core_msq` resume-schema regression guard, adds dedicated reference-backtest alignment coverage, and wraps the fixture-based inline-backtest tests with explicit `pytest.MonkeyPatch()` adapters so they execute under `python -m tests.run` instead of only under pytest fixture injection.

limen/log/_experiment_backtest_results.py
mikkokotila: Addressed in `_experiment_backtest_results`. Regression rounds are now normalized to the same directional binary convention as the inline reference-architecture path before snapshot runs, so post-run backtests stay aligned with inline backtests and the confusion-bucket return columns populate on regression rounds too.

mikkokotila:
Follow-up on the latest review pass: the fresh commit (`3a38938`) closes the remaining inline/post-run alignment gap for regression backtests, adds explicit NaN rejection before inline actuals are cast, adds direct `random_binary` pass-through coverage, harmonizes the xgboost kwargs assertions, and wires those guards into the canonical `python -m tests.run` path. I also reran the canonical suite locally and checked the resulting coverage report: project/package total is 92%, with the touched implementation files at 96%.

zero-bang (approved):

tests/run.py
mikkokotila: Follow-up after `0972bfc`: I intentionally removed the `experiment_core` / MSQ resume-schema guard from this PR. Maintainer decision is that this PR stays limited to snapshot metric additions and their canonical test coverage, without changing mainline `experiment_core` behavior for all-NaN `results.csv` resume edge cases. The runner still covers the metric-surface additions that belong here; any resume-schema handling should be a separate proposal if it is ever needed.

limen/log/_experiment_backtest_results.py
```diff
@@ -5,6 +5,32 @@
...
+
+    actuals = pd.to_numeric(df['actuals'], errors='coerce')
+    valid_actuals = actuals.dropna()
+    if not valid_actuals.empty and valid_actuals.isin([0, 1]).all():
+        return df
+
+    normalized = df.copy()
+    predictions = pd.to_numeric(normalized['predictions'], errors='coerce')
+    normalized['predictions'] = (predictions > 0).astype(int)
+    normalized['actuals'] = (actuals > 0).astype(int)
```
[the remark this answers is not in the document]
mikkokotila: Maintainer decision: not taking this on in this PR. This PR is intentionally limited to adding the snapshot metrics and aligning the binary + directional-regression paths for valid runs. We are not adding extra defensive null-preserving logic here for NaN `actuals` in post-run logs, because that would be complexity in service of a garbage-run / invalid-log anomaly rather than normal Limen behavior. If `actuals` are NaN in this path, that run is already broken upstream; it should be fixed or handled separately, not by expanding this metric PR.

Issue #450, opened 2026-04-19
Upgrade legacy ParamSpace huge-space sampling while preserving exact random.sample semantics


mikkokotila:
## Summary

Legacy `ParamSpace` has a real failure mode on extremely large discrete spaces:

- it calls `random.sample(range(total_space), n_permutations)` when `n_permutations < total_space`
- for sufficiently large `total_space`, that can fail with:
  - `OverflowError: Python int too large to convert to C ssize_t`

We now have the full solution worked out.

This issue is for implementing the huge-space upgrade **without changing legacy semantics**.

## Background

`ParamSpace` is the legacy permutation helper used by the standard non-MSQ `UEL.run()` path.

Important constraints for that path:
- exact seeded reproducibility matters
- selection order matters
- row order in `experiment_log` / streamed CSV matters
- the prefix of a partial run matters
- downstream RNG state after pool construction matters, because `generate(random_search=True)` continues consuming the same module-level `random` state

That means a huge-space fix must preserve the exact effective behavior of:

```python
random.sample(range(N), k)
```

for the `range(N)` case, including selection order and post-call RNG state.

## What was initially proposed, and why it failed

A first attempt was to handle the overflow regime by sampling unique indices into a `set` via repeated `random.randrange(total_space)` calls.

Why that failed:
- it preserved uniqueness
- it approximately preserved without-replacement coverage
- but it did **not** preserve the exact selection order contract once the sampled pool was materialized from an unordered set

Consequences of that failure:
- exact seeded sequence reproducibility breaks
- row identity in `experiment_log` / CSV changes
- the prefix of a partial interrupted legacy run changes

In short: it was statistically acceptable, but semantically wrong for legacy `ParamSpace`.

## Correct engineering concept

The right reframing is:

> We do not need a new sampling algorithm.
> We need to execute the same algorithm that `random.sample(range(N), k)` would have executed in the large-`N`, small-`k` selection branch, but without going through the wrapper path that overflows.

That means:
- keep the existing `random.sample(...)` behavior unchanged when `N` is small enough for the pool branch
- inline the selection branch for huge `N`
- preserve draw order explicitly with an append-ordered result list
- preserve dedup with a `set`
- drive the loop with the same `getrandbits(N.bit_length())` rejection pattern as `_randbelow`

## Implementation shape

Reference implementation shape:

```python
import math


def sample_range_exact(rng, N, k):
    if not 0 <= k <= N:
        raise ValueError("Sample larger than population or is negative")

[...]

PR #452, opened 2026-04-19
feat: Fix huge-space legacy ParamSpace sampling

mikkokotila:
## Summary

Fixes the legacy `ParamSpace` overflow path from [#450](https://github.com/Vaquum/Limen/issues/450) without changing the existing small-space behavior.

This keeps the current `random.sample(range(N), k)` path untouched for representable populations and only swaps in an exact stdlib-equivalent helper when `total_space > sys.maxsize` and the wrapper would otherwise overflow.

## What Changed

- add `sample_range_exact()` in `limen/utils/param_space.py` to mirror CPython's ordered selection-branch semantics for `random.sample(range(N), k)`
- switch `ParamSpace` to use that helper only in the true overflow regime
- harden the helper so its small-range fallback cannot reintroduce the same overflow above `sys.maxsize`
- add focused regression tests for:
  - stdlib-equivalent sampled output and downstream RNG state
  - rejection and dedup behavior in the huge-range helper
  - prefix stability for huge ranges
  - unchanged small legacy `ParamSpace` sequencing
  - no-overflow reproducibility for `19 x 10` and operational coverage beyond that
  - enumerate-all behavior and exhaustion to `None`
  - the guarded no-fallback-above-`sys.maxsize` edge case
- bump the patch version to `2.3.5` and add a changelog entry
- fold the issue comment clarifications back into the issue body so the contract is canonical in one place

## Why

The old implementation failed on extremely large discrete spaces because `random.sample(range(total_space), n_permutations)` eventually trips the `Py_ssize_t` limit in the stdlib wrapper.

The important part here is that legacy `ParamSpace` is sensitive not just to uniqueness, but to:

- selection order
- seeded reproducibility
- downstream RNG state after pool construction
- partial-run prefix identity

So this PR fixes the overflow by preserving those semantics, not by introducing a different sampler.

## Validation

- `source .venv-codex/bin/activate && pytest -q tests/test_param_space.py tests/test_large_param_space.py`
- `source .venv-codex/bin/activate && python -m coverage run -m pytest -q tests/test_param_space.py tests/test_large_param_space.py && python -m coverage report --include='limen/utils/param_space.py'`
- `source .venv-codex/bin/activate && ruff check limen/utils/param_space.py tests/test_param_space.py tests/run.py`
- `source .venv-codex/bin/activate && LIMEN_COVERAGE_RUN=1 python -m coverage run -m tests.run && python -m coverage report --fail-under=92`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -717,3 +717,9 @@ Note: add all new changelog entries to the bottom of this file.
 
 - Align inline and post-run confusion mean-return metrics to the same immediate-next-execution-row contract used by snapshot backtests for completed-bar pipelines
 - Make post-run snapshot backtests raise explicitly on unsupported multiclass logged outputs while keeping the existing binary and directional-regression contracts
+
+## v2.3.5 on 19th of April, 2026
+
+- Fix legacy `ParamSpace` sampling for huge discrete spaces so `n_permutations < total_space` no longer overflows when `total_space > sys.maxsize`
+- Preserve exact legacy `random.sample(range(N), k)` ordering and downstream RNG-state semantics in the huge-space fallback path
+- Add regression coverage for stdlib-equivalent range sampling, large-space reproducibility, prefix stability, and exhaustion behavior on the legacy `ParamSpace` path
--- a/limen/utils/param_space.py
+++ b/limen/utils/param_space.py
@@ -44,7 +44,7 @@ def sample_range_exact(
             )
         )
 
-    if population_size <= setsize:
+    if population_size <= setsize and population_size <= sys.maxsize:
         return rng.sample(range(population_size), sample_size)
 
     bits = population_size.bit_length()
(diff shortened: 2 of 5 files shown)


zero-bang (approved):
Clean, correct implementation. Verified:

- The constants (21, 5, 4, 3) and branch logic exactly mirror CPython `Lib/random.py`'s selection-sampling path for `range(N)` objects.
- The combined rejection loop (`>= N or in seen`) produces identical `getrandbits` call sequences to CPython's separate `_randbelow` + duplicate-rejection loops — confirmed across 280 (seed, N, k) cases with zero mismatches.
- The `sys.maxsize` boundary is the correct cutoff: `len(range(sys.maxsize))` succeeds, `len(range(sys.maxsize + 1))` raises `OverflowError`.
- All 9 tests pass and cover the right surface: stdlib equivalence, rejection/dedup ordering, edge cases, prefix stability, legacy backward compat, exhaustion, reproducibility for both generation modes.
- Callers checked: only `experiment_core.py` (legacy non-MSQ path) and `tests/test_large_param_space.py` — no missed consumers.

pdey (approved):

PR #453, opened 2026-04-19
Move mean return metrics to confusion side

mikkokotila:
## Summary
- move the TP/FP/TN/FN mean-return fields off the snapshot backtest surface and onto the confusion side where they semantically belong
- restore `backtest_snapshot()` and inline snapshot backtests to price-only inputs with no `actuals` plumbing
- add aligned next-bar confusion-return support post-run through `experiment_confusion_metrics('aligned_return_pct')` and inline through `confusion_*_mean_return_pct`

## Scope Note
- this PR is intentionally limited to the confusion/backtest metric boundary
- it does **not** change `experiment_core` or mainline log/resume behavior
- it keeps the latest merged snapshot execution-lag and run-level trade semantics intact

## Implementation Notes
- inline experiment rows now expose `confusion_tp_mean_return_pct`, `confusion_fp_mean_return_pct`, `confusion_tn_mean_return_pct`, and `confusion_fn_mean_return_pct`
- post-run confusion analysis now supports `aligned_return_pct`, including `tp/fp/tn/fn_mean_return_pct` aliases on that surface
- `experiment_backtest_results()` keeps the trading metrics only; confusion-conditioned return means are no longer emitted from the backtest table

## Testing
- `uv run --with pytest pytest tests/test_backtest_snapshot.py tests/test_reference_architecture.py tests/test_reference_backtest_alignment.py tests/test_metrics_and_log_helpers.py tests/test_inline_metrics.py -q`
- `uv run --with ruff ruff check tests/run.py limen/backtest/backtest_snapshot.py limen/log/_experiment_backtest_results.py limen/log/_permutation_confusion_metrics.py limen/sfd/reference_architecture/base.py limen/sfd/reference_architecture/logreg_binary.py limen/sfd/reference_architecture/random_binary.py limen/sfd/reference_architecture/tabpfn_binary.py limen/sfd/reference_architecture/xgboost_regressor.py tests/test_backtest_snapshot.py tests/test_reference_architecture.py tests/test_reference_backtest_alignment.py tests/test_metrics_and_log_helpers.py tests/test_inline_metrics.py`
- `uv run --with pytest --with coverage coverage run -m tests.run`
- `uv run --with coverage coverage report` (`TOTAL 93%`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -687,13 +687,19 @@ Note: add all new changelog entries to the bottom of this file.
 - Add cross-timescale state features: `trend_coherence` and `volatility_term_structure`
 - Add documentation and full behavioral test coverage for the new OHLCV feature families
 
+## v2.3.1 on 19th of April, 2026
+
+- Move TP/FP/TN/FN mean return metrics off the snapshot backtest surface and onto the confusion/benchmark side
+- Add inline `confusion_tp_mean_return_pct`, `confusion_fp_mean_return_pct`, `confusion_tn_mean_return_pct`, and `confusion_fn_mean_return_pct`
+- Add `aligned_return_pct` support to post-run confusion metrics so the same next-bar aligned one-bar return view is available after a run
+- Restore `backtest_snapshot()` and inline snapshot backtests to price-only inputs with no label plumbing through the backtest helper
+
 ## v2.3.0 on 18th of April, 2026
 
 - Change snapshot backtests to execute predictions on the next tradable bar by default, with `execution_lag_bars=0` available for legacy same-row behavior
 - Change snapshot `trade_*` metrics to run-level by default, add explicit `bar_*` metrics, and retain `trades_count_mode='bars'` for legacy-style bar metrics
 - Apply the new snapshot defaults explicitly to experiment backtest summaries and SFD inline backtest metrics so the behavior change is intentional and visible in outputs
-- Add confusion-bucket mean return columns (`tp/fp/tn/fn_mean_return_pct`) to snapshot backtest summaries when aligned `actuals` are available
-- Populate inline snapshot confusion-bucket return metrics from the aligned test labels, with regression SFDs passing directional actuals explicitly
+- Add confusion-bucket mean return diagnostics for aligned next-bar returns
 - Align regression experiment/log snapshot backtests with the same directional binary convention as inline reference-architecture backtests
 - Add `mean_kelly_pct` to snapshot backtest summaries using the active trade/bar return distribution with breakeven observations kept in the Kelly sample
 - Add direct snapshot tests to the canonical `python -m tests.run` path and extend docs for the updated snapshot semantics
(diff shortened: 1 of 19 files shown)


limen/backtest/backtest_snapshot.py
```diff
@@ -71,6 +68,7 @@ def backtest_snapshot(df: pd.DataFrame,
     if execution_lag_bars < 0:
         raise ValueError('execution_lag_bars must be >= 0')
 
+    _ = actual_col  # kept for call-site compatibility after confusion returns moved out of snapshot
```
mikkokotila: We just added this in the previous PR, so I don't think we should keep it here.

CHANGELOG.md
```diff
@@ -687,13 +687,19 @@ Note: add all new changelog entries to the bottom of this file.
 - Add cross-timescale state features: `trend_coherence` and `volatility_term_structure`
 - Add documentation and full behavioral test coverage for the new OHLCV feature families
 
+## v2.3.1 on 19th of April, 2026
```
zero-bang, line 690: v2.3.1 is placed **above** v2.3.0, but line 3 of this file says "add all new changelog entries to the bottom of this file." The rest of the file follows ascending order. Move the v2.3.1 block below the v2.3.0 block.

limen/sfd/reference_architecture/base.py
```diff
@@ -100,21 +100,83 @@ def _compute_confusion(self, preds: np.ndarray, y_test: np.ndarray) -> dict:
...
+        dpx = close_px - open_px
+
+        if execution_lag_bars > 0:
+            open_px = open_px.shift(-execution_lag_bars)
+            dpx = dpx.shift(-execution_lag_bars)
+
+        aligned_return_pct = ((dpx / open_px).replace([np.inf, -np.inf], np.nan) * 100.0)
+        valid = open_px.notna() & dpx.notna() & (open_px != 0) & aligned_return_pct.notna()
+
+        actual = pd.Series(y_test)
```
zero-bang, line 157: `actual` and `pred` are created with a default `RangeIndex`, but `valid` and `aligned_return_pct` inherit whatever index `price_pd` has. If `price_pd` has a non-default index (e.g. a `DatetimeIndex` from a raw pandas caller), the `&` operations silently produce all-`False` masks because the indices don't align, and every metric comes back `NaN`.

The adjacent `_compute_backtest` method already handles this correctly by extracting `.values` (line 205). Easiest fix here is to reset the index after conversion:

```python
price_pd = price_pd.reset_index(drop=True)
```

right after the `to_pandas()` / isinstance block (around line 135). This makes the method robust regardless of caller.

zero-bang (changes requested):
Clean refactor with good semantic separation — confusion-conditioned returns belong on the confusion side, not in the backtest output. Tests are comprehensive, CI is green, and the scope is well-bounded.

Two findings below. The CHANGELOG ordering is a quick fix; the index alignment is a latent correctness issue that doesn't manifest today but would bite silently if `price_data_for_backtest` is ever a pandas DataFrame with a non-default index.

PR #459, opened 2026-04-19
fix: make snapshot execution yield alignment right

mikkokotila:
## What changed
- adds an explicit `execution_lag_bars=1` alignment to snapshot backtests in the inline and post-run snapshot paths
- preserves the existing HOLD-WHILE-1 mechanics once a prediction reaches a tradable execution row
- adds deterministic proof tests at three levels: direct snapshot behavior, generic manifest/data preparation, and the real `logreg_binary` manifest path
- bumps the release metadata to `2.3.3`

## Why
The current completed-bar pipelines build features on the finished feature row while `y_test` is already shifted into the future. Snapshot backtests were still pairing those predictions with the same row's price move. This PR narrows the change to just the execution alignment so we can validate the economic effect without bringing back the broader surface from `#447`.

## Proof / validation
- `uv run --with pytest pytest tests/test_manifest_prepare_data.py tests/test_metrics_and_log_helpers.py tests/test_inline_metrics.py tests/test_reference_architecture.py -q`
- full `uv run --with pytest --with coverage coverage run -m tests.run` is still running locally while this PR goes up for review

## Notes
- this PR intentionally does not touch `experiment_core`
- this PR intentionally preserves HOLD-WHILE-1 semantics after the execution-row shift


zero-bang (approved):
Thoroughly reviewed. The execution lag shift logic is correct — I traced through every test case by hand and confirmed the shift/eval_mask/pos pipeline produces the right positions. HOLD-WHILE-1 continuation across the lag boundary works because the first execution bar can only be an entry (pos at bar lag-1 is always False due to eval_mask), and subsequent bars correctly fall into cont_mask.

All three callers (two production, one pre-existing test) are updated: the two production paths pass execution_lag_bars=1 explicitly, and the existing kelly test now passes execution_lag_bars=0 to preserve its original semantics. The new default of 1 is appropriate for the completed-bar pipeline.

The bars_total semantic change from len(df) to eval_mask.sum() is more correct — pre-lag bars are untradable and shouldn't inflate the denominator. The Sharpe change to use eval_returns (R_net filtered by eval_mask) is similarly correct.

Tests are meaningful: the _DummyCompletedBarSignal proof test is the strongest — it shows the same data yields +21% with same-row pricing (look-ahead bias) and -19% with next-bar pricing (correct alignment). The manifest-level tests verify price/feature row alignment through the real pipeline.

LGTM.

tests/test_metrics_and_log_helpers.py
```diff
@@ -171,13 +197,87 @@ def test_backtest_snapshot_adds_mean_kelly_pct() -> None:
...
+    assert result['trades_count'] == 1
+    assert result['total_return_gross_pct'] == 21.0
+    assert result['total_return_net_pct'] == 21.0
+
+
+def test_backtest_snapshot_excludes_predictions_without_next_tradable_bar() -> None:
+    result = backtest_snapshot(
+        pd.DataFrame({
+            'predictions': [1, 0, 0],
+            'open': [100.0, np.nan, 100.0],
```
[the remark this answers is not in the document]
mikkokotila: Renamed the test to `test_backtest_snapshot_drops_predictions_without_immediate_next_execution_bar` so it reflects the actual contract: we do not skip ahead to a later tradable row.

CHANGELOG.md
```diff
@@ -706,3 +706,8 @@ Note: add all new changelog entries to the bottom of this file.
 
 - Add `mean_kelly_pct` back to snapshot backtests without changing the restored legacy execution semantics
 - Add TP/FP/TN/FN mean return percentage columns to the existing confusion metrics surfaces, both inline and post-run, using the existing aligned `open` and `price_change` data
+
+## v2.3.3 on 19th of April, 2026
+
+- Fix `backtest_snapshot()` execution alignment for completed-bar pipelines by evaluating prediction row `t` on the next tradable bar while preserving the existing HOLD-WHILE-1 semantics
+- Apply the same snapshot execution alignment explicitly to inline reference-architecture backtests and post-run experiment backtest summaries
```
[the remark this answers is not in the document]
mikkokotila: Clarified the changelog wording to say "immediate next execution row" so it matches the implemented contract exactly. This is a wording-only follow-up; behavior is unchanged.

mikkokotila (commented):
## Review — focus on the shift, whole-Limen picture

Read the full diff; traced the data flow end-to-end from `Manifest.prepare_data` through `permutation_prediction_performance` into both the `backtest_snapshot` path and the confusion-bucket path; re-ran the PR's tests and a concrete reproducer on the new branch. **All 34 tests pass.** The shift inside `backtest_snapshot` is correct. There is one critical whole-Limen inconsistency that this PR does not close, described in detail below.

### 1. The shift in `backtest_snapshot` is correct

I want to record the reasoning so future reviewers can re-verify without re-deriving.

The pipeline contract, verified against `limen/experiment/manifest_core.py` and `limen/log/_permutation_prediction_performance.py`:

- `price_data_for_backtest` is **on the feature bar** — `test_split.select(['datetime','open','high','low','close'])` is taken before any transforms ([manifest_core.py:634-637](limen/experiment/manifest_core.py#L634)), then after transforms + `drop_nulls` the final datetimes are left-joined back onto it ([manifest_core.py:701-706](limen/experiment/manifest_core.py#L701)). So `price_data_for_backtest[t].open/close` are bar t's own prices.
- `y_test` **is already shifted forward** by the target transform (e.g. `shift_column_transform(shift=-1, column='target_column')` in the logreg SFD family). So `y_test[t]` describes bar t+1's direction.
- `permutation_prediction_performance` pairs `predictions[t]` (built from bar t's features) with `open[t]/close[t]/price_change[t]` (bar t's prices) and `actuals[t]` (shifted label about bar t+N).
- [test_shifted_target_pipeline_keeps_price_rows_on_feature_bar](tests/test_manifest_prepare_data.py) locks this contract: `current_sign == price_sign == [1,0,1]` (feature-bar), `shifted_target == [0,1,0]` (label about the next bar).

Given that contract, shifting predictions forward by 1 inside `backtest_snapshot`:

```python
pred = pred.shift(execution_lag_bars, fill_value=0)
eval_mask = execution_rows & tradable
pos = (pred == 1) & eval_mask
```

is exactly the right move. At row t, `pos[t] = (pred_original[t-1] == 1) & tradable[t]`, so `r_entry[t] = (close[t]-open[t])/open[t]` is bar t's return, triggered by a decision observed at bar t-1's close. Fully causal.

I also verified there is **no double-shift**: the prep pipeline only shifts `y_test` (the target column). Prices in `price_data_for_backtest` and in `_get_test_data_with_all_cols` are untouched. Snapshot is the only place the prediction-vs-price pairing is re-aligned, and it's applied exactly once in both surfaces that matter ([base.py:150](limen/sfd/reference_architecture/base.py#L150) and [_experiment_backtest_results.py:29-32](limen/log/_experiment_backtest_results.py#L29)).

The pivotal test [test_completed_bar_signal_proves_next_bar_alignment](tests/test_metrics_and_log_helpers.py) makes this watertight: a perfect-look-ahead signal shows `+21.0%` with `execution_lag_bars=0` and `-19.0%` with `execution_lag_bars=1`. That 40-point gap is the size of the look-ahead bias that was silently bleeding into every legacy backtest run on completed-bar pipelines. This is the most important test in this PR; keep it pinned.

### 2. Critical whole-Limen gap: confusion-bucket mean returns are NOT shifted

This is the main thing to fix, and it's the one the user asked me to hunt for. The shift is right inside `backtest_snapshot`, but the **exact same look-ahead bias the PR just eliminated from the backtest is still present in the confusion-bucket mean-return metrics that sit next to the backtest in every output dict.**

The two surfaces now disagree.

**Where the unshifted code lives:** [`_confusion_mean_return_pct` in limen/log/_permutation_confusion_metrics.py:30](limen/log/_permutation_confusion_metrics.py#L30) — `return_pct = (price_change / open_px) * 100.0` pairs `pred[t]` with bar t's return, no shift applied.

**Who calls it with same-row inputs:**
- [`ReferenceModel._compute_confusion`](limen/sfd/reference_architecture/base.py#L115) — emits `confusion_tp_mean_return_pct` etc. inline, alongside `backtest_*` keys, in `evaluate()`'s return dict.
- [`_permutation_confusion_metrics`](limen/log/_permutation_confusion_metrics.py#L188) — post-run, uses `df['open']`/`df['price_change']` from `permutation_prediction_performance`, so also on the feature bar.
- [`UniversalExperimentLoop.experiment_confusion_metrics`](limen/experiment/experiment_core.py#L278) — aggregates the post-run version per round.

**Reproducer** (I ran this on the PR branch):

```python
# perfect-look-ahead signal with shift=-1 label convention
preds   = np.array([1, 0, 1, 0])
actuals = np.array([0, 1, 0, 1])    # next-bar direction
open_   = np.array([100.]*4)
close_  = np.array([110., 90., 110., 90.])
pc      = close_ - open_
perf = pd.DataFrame({'predictions': preds, 'actuals': actuals,
                     'open': open_, 'close': close_, 'price_change': pc})

_confusion_mean_return_pct(perf['predictions'], perf['actuals'],
                           perf['open'], perf['price_change'])
# → {'tp_mean_return_pct': nan, 'fp_mean_return_pct': 10.0,  ← LOOK-AHEAD BIAS
#    'tn_mean_return_pct': nan, 'fn_mean_return_pct': -10.0}

backtest_snapshot(perf, execution_lag_bars=1, fee_bps=0, slip_bps=0).iloc[0]
# → total_return_net_pct = -19.0                             ← CAUSAL
```

Same frame. Same round. One surface says "the FP bucket averaged `+10%` per bar" (look-ahead), the other says "the strategy lost `19%` net" (causal). Both end up in the same `evaluate()` return dict. A user reading both columns side-by-side will draw contradictory conclusions about the same signal.

When I shift `price_change`/`open` backward by 1 (mirroring what `backtest_snapshot` now does to `pred`), the confusion numbers flip sign and match the backtest:

```
cb_nextbar = {'tp_mean_return_pct': nan, 'fp_mean_return_pct': -10.0,
              'tn_mean_return_pct': nan, 'fn_mean_return_pct': 10.0}
```

This is the number the user actually wants — **the mean of the executed bar's return for rows that fell into each confusion bucket**, which is what `tp/fp/tn/fn_mean_return_pct` is documented to be in `docs/Backtest.md` from the v2.3.0 entry.

**Severity:** High. The whole point of the PR is to remove look-ahead from the metrics Limen surfaces to the user. Doing it only on the `backtest_*` side and leaving it intact on the `confusion_*` side means the look-ahead bias has moved from being a disaster everyone saw to being a disaster only the person reading one specific column sees. If anything that is worse than the original problem, because it silently undermines the credibility of the newly-correct `backtest_*` numbers.

**Fix.** Either of:

1. **Preferred — apply the shift inside `_confusion_mean_return_pct` consistently.** Add an `execution_lag_bars: int = 1` parameter that shifts `open_px` and `price_change` backward by that many rows (or equivalently shifts `pred`/`actual` forward) and drops the first `execution_lag_bars` rows from the mean. Default to 1 so every caller (`_compute_confusion`, `_permutation_confusion_metrics`, the aggregator in `experiment_core`) inherits the correct semantics without needing to opt in. Add a test modelled on `test_completed_bar_signal_proves_next_bar_alignment`:

    ```python
    # perfect look-ahead signal on completed-bar pipeline
    same_row = _confusion_mean_return_pct(..., execution_lag_bars=0)
    next_bar = _confusion_mean_return_pct(..., execution_lag_bars=1)
    assert same_row['fp_mean_return_pct'] == 10.0   # look-ahead
    assert next_bar['fp_mean_return_pct'] == -10.0  # causal
    ```

2. **Fallback — rename and document.** If the confusion-bucket mean returns are deliberately supposed to express "the mean price move of the *feature* bar for each quadrant" (a diagnostic of the signal's self-consistency, not of its tradable edge), then rename the columns (e.g. `confusion_tp_feature_bar_return_pct`) and say so loudly. As things stand today, `tp_mean_return_pct` appearing next to `backtest_total_return_net_pct` will be read by users as "the per-bar return that backs up the backtest number" — and it isn't.

I'd strongly prefer option 1. The confusion metrics and the backtest should be telling the same story. If they don't, one of them is wrong, and right now the confusion side is wrong for completed-bar pipelines.

### 3. Secondary observations (not blockers)

- **`bars_total` semantic change.** Now `bars_total = int(eval_mask.sum())` instead of `len(df)` — subtracts the trailing `execution_lag_bars` rows plus any non-tradable rows. Correct in spirit, but a silent downstream change for anyone computing ratios with `bars_total` as a denominator. One line in the changelog would cover it.

- **`execution_lag_bars=0` still has the new `tradable` masking.** Fine for most inputs (all rows tradable → no change), but if someone runs with `execution_lag_bars=0` on a frame with NaN OHLC rows, `bars_total` and `bars_in_market_pct` will differ vs. the legacy build even though "legacy same-row" is the advertised behavior for that flag. Minor — probably intended — but worth an explicit note.

- **Dead-variable warning.** [backtest_snapshot.py:108](limen/backtest/backtest_snapshot.py#L108) keeps the pre-existing `_trade_return_mean_pct = np.nan` in the else branch. Never read. Safe to delete next pass.

- **Test coverage gap.** `test_inline_and_post_experiment_metrics` now asserts `backtest_total_return_net_pct` inline == post-run — excellent. But there is no matching assertion that `confusion_tp_mean_return_pct` inline == post-run, and no assertion that any `confusion_*_mean_return_pct` agrees with the corresponding `backtest_*` sign on a look-ahead reproducer. Worth adding once the fix in §2 lands, so the two surfaces are anchored together going forward.

- **[_experiment_backtest_results.py:38](limen/log/_experiment_backtest_results.py#L38) trailing newline removed.** No-op stylistically, but unrelated to the shift — fine to keep out of the review narrative.

### 4. Verdict

The shift in `backtest_snapshot` is **correct and verifiably causal** for completed-bar pipelines. The tests nail the contract at the snapshot level, and the pipeline invariants they rely on (`price_data_for_backtest` on the feature bar, label pre-shifted, no double-shift elsewhere) all check out when I trace them.

But if the user's stated bar is *"if we have issue in the shift, we have issue in everything"*, then this PR closes half the surface. The confusion-bucket mean-return computation is the other half — same inputs, same alignment concern, same look-ahead bias, but none of the fix. Please extend the same `execution_lag_bars=1` treatment to [_confusion_mean_return_pct](limen/log/_permutation_confusion_metrics.py#L30) and its three call sites before merge. With that, the whole Limen picture is consistent.

tests/test_manifest_prepare_data.py
```diff
@@ -74,6 +89,80 @@ def test_price_data_for_backtest_datetime_alignment() -> None:
...
+    })
+
+    data = manifest.prepare_data(raw_data, {'bar_type': 'base'})
+    price_df = data['price_data_for_backtest']
+
+    current_sign = data['x_test']['current_sign'].to_list()
+    price_sign = (price_df['close'] > price_df['open']).cast(pl.Int8).to_list()
+    shifted_target = data['y_test'].to_list()
+
+    assert current_sign == [1, 0, 1]
```
mikkokotila, line 116: [P1] This codifies the feature-bar alignment, but the tail of that contract is still wrong for a next-execution-row backtest. Because `price_data_for_backtest` is trimmed down to the same retained datetimes as `x_test`, the final kept prediction has no slot for its real next raw bar even when that bar exists. I reproduced this against the branch with this exact `shift=-1` toy manifest: the raw test window still contains `2025-01-01 07:00`, but `_alignment["last_test_datetime"]` stops at `06:00`, so a positive prediction on the `06:00` feature row is dropped instead of being priced on `07:00`. If row `t` is supposed to execute on the immediate next execution row, the snapshot input needs to preserve that tail execution row somewhere.

limen/log/_experiment_backtest_results.py
```diff
@@ -26,11 +26,13 @@ def _experiment_backtest_results(self: Any, disable_progress_bar: bool = False)
 
     for i in tqdm.tqdm(range(len(self.round_params)), disable=disable_progress_bar):
 
-        result_df = backtest_snapshot(self.permutation_prediction_performance(i))
+        result_df = backtest_snapshot(
```
mikkokotila, line 53: [P1] Logged regression backtests still disagree with the inline path here. This now feeds raw `permutation_prediction_performance()` output straight into `backtest_snapshot()`, but snapshot coerces predictions with `.astype(int).clip(0, 1)`, while `XGBoostRegressor.evaluate()` directionalizes with `preds > 0`. So a logged regression prediction like `0.4` becomes `0` here but `1` inline. That means the same regression round can still show different economics depending on whether the user reads `uel.experiment_backtest_results` or the inline backtest metrics.

tests/test_metrics_and_log_helpers.py
```diff
@@ -171,13 +197,87 @@ def test_backtest_snapshot_adds_mean_kelly_pct() -> None:
...
         }),
+        execution_lag_bars=0,
         fee_bps=0.0,
         slip_bps=0.0,
     ).iloc[0]
 
     assert result['mean_kelly_pct'] == 25.0
 
 
+def test_backtest_snapshot_executes_on_next_bar() -> None:
```
mikkokotila, line 225: [P2] This is the right proof case, but it still is not on the canonical CI path the repo actually runs. GitHub Actions executes `python -m coverage run -m tests.run`, and `tests/run.py` still does not import or call this test or the other new shift-contract cases added in this PR. So the semantics we are trying to lock down here are still not protected by the suite CI treats as authoritative.

mikkokotila (commented):
Focused another pass on the shift contract end-to-end. These three comments are the remaining places where I think the fix is still incomplete.

tests/test_manifest_prepare_data.py
mikkokotila: Leaving this one as an explicit maintainer decision for this PR: we are not widening the `price_data_for_backtest` contract or preserving an extra tail execution row here. For the completed-bar fix in `#459`, we accept that there is one fewer executable row at the tail and keep the snapshot input shape unchanged. The broader tail-row semantics question is real, but it is outside the blast radius we are allowing in this PR.

limen/log/_experiment_backtest_results.py
mikkokotila: Fixed on the new head. Post-run snapshot summaries now normalize logged non-binary predictions with the same directional contract the inline regression path already uses (`preds > 0`), instead of relying on snapshot's integer cast. The change is kept local to `_experiment_backtest_results.py`, and `test_experiment_backtest_results_directionalizes_regression_predictions` now locks the inline/post-run parity for this case.

tests/test_metrics_and_log_helpers.py
mikkokotila: Wired the direct shift-contract coverage into the canonical runner on the new head. `tests.run` now imports and executes the direct snapshot alignment proofs, the manifest feature-bar alignment proofs, and the new regression parity check, so the authoritative `python -m coverage run -m tests.run` path now covers the semantics this PR is introducing.

zero-bang (approved):

zero-bang (approved):

mikkokotila (commented):
Re-review on `5bde2c9` follow-up notes. The shift itself is correct end-to-end and `test_completed_bar_signal_proves_next_bar_alignment` is the right anchor. The confusion-surface gap I raised earlier is being handled in a separate PR per the author — tracking it out of scope here. A few smaller things worth picking up before merge:

### Regression directionalization — two edge cases in `_prepare_snapshot_backtest_input`

The helper correctly normalizes binary preds (passed through) and regression preds (directionalized via `(pred > 0).astype(int)`). Two input shapes silently do the wrong thing:

1. **Probabilities fed directly as preds.** A custom SFD that stores `[0.6, 0.3, 0.8]` (probabilities, not directions) silently becomes `[1, 1, 1]` — all positive probabilities are `> 0`. No existing reference SFD does this, but nothing in the helper prevents it, and the resulting backtest will look spuriously long-biased. Suggest a short docstring line: "Assumes non-binary predictions are directional scores whose sign maps to the long/flat decision, not probabilities."

2. **Ternary `[1, 0, -1]` short signals.** Collapse to `[1, 0, 0]`. Correct for today's long-only backtest, but if short support lands later, this helper will silently strip the short side. Fine for now; worth a comment pinning the long-only assumption so the eventual short-enabling change knows where to look.

### Coverage gap — regression inline vs post-run agreement

`test_inline_and_post_experiment_metrics` asserts `backtest_total_return_net_pct` inline == post-run using `LogRegBinary` (binary). The two directionalization code paths for regression — inline `(preds > 0).astype(int)` in `XGBoostRegressor` and post-run `_prepare_snapshot_backtest_input` — must agree end-to-end, but no test runs `XGBoostRegressor` through the actual UEL and compares the columns. Controlled manual check confirms they agree at default fees (`total_return_net_pct = -19.4` both sides for the same preds), but it isn't anchored in the suite. A copy of the existing test swapping `LogRegBinary` for `XGBoostRegressor` would close it cheaply.

### Missing changelog line — `bars_total` semantics changed

`bars_total` in `backtest_snapshot` now equals `int(eval_mask.sum())` (tradable execution rows) rather than `len(df)`. For completed-bar pipelines at `execution_lag_bars=1` this subtracts at least one row, plus any non-tradable rows, from the value users previously saw. The current v2.3.3 entry doesn't mention it. One line in CHANGELOG.md would cover it, e.g.:

> - `bars_total`, `bars_in_market_pct`, and `sharpe_per_bar` in `backtest_snapshot` are now computed on the tradable window, so the trailing `execution_lag_bars` rows and any non-tradable rows are excluded from denominators. Historical tables using these columns are not directly comparable across the upgrade.

### Missing doc entry — `execution_lag_bars=0` is no longer a pure legacy fallback

`docs/Backtest.md` says `execution_lag_bars=0` reproduces the legacy same-row behavior. Almost true — but the new `tradable = open_px.notna() & close_px.notna() & dpx.notna() & (open_px != 0)` and the `bars_total = int(eval_mask.sum())` definition both apply even when `execution_lag_bars=0`. A frame with NaN OHLC rows run through `execution_lag_bars=0` now returns a different `bars_total` / `bars_in_market_pct` / `sharpe_per_bar` than the pre-PR same-row code did. The diff is tiny in practice but the doc should say "same-row execution of tradable rows" rather than "same-row execution," so nobody reading the doc is surprised when a NaN-bearing fixture yields a different `bars_total` under the new `execution_lag_bars=0`.

### Tiny — dead variable

`backtest_snapshot.py:108` has `_trade_return_mean_pct = np.nan` inside a chained assignment in the else branch; the name is never read anywhere. Leftover from a refactor. One-line delete.

mikkokotila (commented):
Splitting the two halves of the coverage-gap note from above per offline discussion:

### Minimal — propose including in this PR

`test_inline_and_post_experiment_metrics` already anchors `backtest_total_return_net_pct` agreement inline vs post-run. The same invariant should hold for the confusion-bucket mean returns — they flow from `_compute_confusion` (inline) and `_permutation_confusion_metrics` (post-run), both of which call the same `_confusion_mean_return_pct` helper, so they must produce identical values for the same round. One extra block anchors it:

```python
for col in ['tp_mean_return_pct', 'fp_mean_return_pct',
            'tn_mean_return_pct', 'fn_mean_return_pct']:
    assert uel.experiment_log[f'confusion_{col}'].tolist() == \
        uel.experiment_confusion_metrics[col].tolist()
```

This is pure regression coverage — it wouldn't change behavior, just pins the invariant so a future refactor can't silently break it. Trivial to fold into this PR.

### Regression inline vs post-run end-to-end

Tracking that one separately as an issue so it doesn't expand this PR's scope: #462.

Issue #456, opened 2026-04-19
Umbrella: clean up log/results semantics and contracts


mikkokotila:
## Purpose
This issue is now the umbrella for Limen's log/results cleanup line.

Originally `#456` tracked the snapshot execution-alignment fix itself. That base fix landed in PR #459. The remaining semantics cleanup is tracked here so follow-up work stays explicit instead of being scattered across PR comments.

## Scope of this umbrella
This is specifically about the log/results side of Limen:
- inline vs post-run agreement
- confusion vs backtest agreement
- snapshot/log input normalization contracts
- wording/docs for those contracts

Not in scope here:
- broader UEL / Manifest cleanup
- sequential backtest redesign
- Explorer UX
- unrelated historical-data or SFD expansion work

## Base fix already landed
- [x] Snapshot backtest execution alignment for completed-bar pipelines landed in PR #459.

## Active follow-ups
- [ ] #463 Finalize remaining log/results semantics: confusion-return execution alignment and explicit snapshot input contract
- [ ] #465 Preserve true next-bar confusion alignment through outlier filtering
- [ ] #466 Handle probability-style snapshot inputs explicitly

## Connected work already completed in this cleanup line
- PR #459 resolved the regression inline/post-run normalization gap originally tracked in #455.
- PR #459 also added the end-to-end regression inline-vs-post agreement test that #462 asked for.

## Older related background
These remain useful context, but they are not the current active cleanup tickets:
- #212 fix: make backtest table returns % correct
- #172 Improve `uel.log` UX
- #117 Unify `results_df` to work with all SFMs

## Folded tickets
The following tickets were previously folded into #463 so the remaining semantics cleanup stayed concentrated:
- #460 Align confusion-return metrics with snapshot execution semantics
- #451 Handle multiclass rounds explicitly in `_prepare_snapshot_backtest_input`

## Done when
We can call this cleanup line complete when:
- confusion-surface returns and backtest returns tell the same economic story
- snapshot/log input normalization has explicit, tested contracts for binary, directional-regression, and multiclass cases
- post-run confusion-return metrics preserve true next-bar alignment even when outlier filtering removes intermediate rows
- docs describe the actual production contracts users see in the tables

mikkokotila:
Reviewed against the current `backtest_snapshot()` and `_permutation_prediction_performance()` implementations. The description is technically correct:

- `_permutation_prediction_performance` stores `pred[t]` alongside row `t`'s own `open`/`close`/`price_change`, and
[...]

PR #464, opened 2026-04-19
Finalize remaining log/results semantics

### LINKED ISSUE #463: Finalize remaining log/results semantics: confusion-return execution alignment and explicit snapshot input contract
## Summary
This issue consolidates the remaining log/results semantics cleanup after PR #459.

It folds together:
- #460 Align confusion-return metrics with snapshot execution semantics
- #451 Handle multiclass rounds explicitly in `_prepare_snapshot_backtest_input`

The goal is to finish the remaining semantics cleanup in one clinical PR, with minimal blast radius and no new parallel surfaces.

## Current state on `main`
Two remaining mismatches are still present in the production code:

1. Confusion-return metrics and snapshot backtests do not use the same economic timing.
   - `limen/backtest/backtest_snapshot.py` now evaluates completed-bar pipelines on the immediate next execution row by default (`execution_lag_bars=1`).
   - But `limen/log/_permutation_confusion_metrics.py::_confusion_mean_return_pct()` still computes:

   ```python
   return_pct = (price_change / open_px) * 100.0
   ```

   on the same row supplied by:
   - inline path: `limen/sfd/reference_architecture/base.py::ReferenceModel._compute_confusion(...)`
   - post-run path: `limen/log/_permutation_confusion_metrics.py::_permutation_confusion_metrics(...)`

   So the confusion-return surface still tells a same-row economic story while snapshot backtests tell a next-execution-row story.

2. `_prepare_snapshot_backtest_input()` still treats every non-binary prediction series as directional long/flat by sign.
   - That is correct for today's directional-regression path.
   - It is not explicit enough for multiclass rounds.
   - Multiclass values are still silently collapsed instead of being rejected with a clear unsupported contract.

## Why this should be fixed together
These are two parts of the same remaining contract problem:
- what economic row a logged prediction is evaluated on
- what prediction types the snapshot/log pipeline explicitly supports

If we fix only one side, the log/results layer still stays ambiguous.

## Minimal fix target
This should be solved in one small semantics PR with these boundaries:

### In scope
- make confusion-return metrics use the same execution semantics as snapshot backtests
- make snapshot input normalization explicit for:
  - binary 0/1 predictions
  - directional regression scores (sign -> long/flat)
  - unsupported multiclass inputs
- tighten wording/docs so the production contract is stated exactly as implemented
- add direct tests that prove the contract end-to-end

### Out of scope
- no `experiment_core` changes
- no `manifest_core` redesign
- no sequential backtest changes
- no new result surfaces
- no short-side support
- no probability heuristics or extra auto-detection beyond the explicit contract

## Code-path plan
Study basis: current `main` implementations in:
- `limen/log/_permutation_confusion_metrics.py`
- `limen/log/_experiment_backtest_results.py`
- `docs/Backtest.md`
- `tests/test_metrics_and_log_helpers.py`
- `tests/test_inline_metrics.py`

### Planned implementation
1. Leave `backtest_snapshot()` unchanged.
   - PR #459 already established the execution contract there.
   - This follow-up should align the remaining log/results surfaces to that contract, not reopen it.

2. Keep the confusion work inside the existing confusion path.
   - Reuse `_confusion_mean_return_pct()` rather than inventing a new surface.
   - Extend that helper minimally so it can compute confusion-quadrant mean returns on the executed row contract instead of same-row `price_change / open`.

3. Use that same helper in the two places that already own confusion outputs.
   - post-run: `_permutation_confusion_metrics(...)`
   - inline: `ReferenceModel._compute_confusion(...)`

4. Keep `_prepare_snapshot_backtest_input()` narrow and explicit.
   - binary `0/1`: pass through
   - directional regression scores: normalize with sign (`pred > 0`)
   - multiclass: raise clearly as unsupported for snapshot/log backtests today

5. Tighten wording in docs/comments only where the contract is user-visible.
   - clarify confusion-return timing once aligned
   - clarify snapshot input normalization support/limits
   - avoid broader docs churn

6. Prove it with direct tests.
   - same toy round must yield matching inline and post-run confusion-return metrics
   - confusion-return metrics must align with the shifted execution contract rather than feature-bar returns
   - multiclass input to `_prepare_snapshot_backtest_input()` must fail clearly
   - binary and directional-regression behavior must remain intact

## Acceptance criteria
- `confusion_*_mean_return_pct` tells the same economic story inline and post-run
- confusion-return timing no longer contradicts `backtest_snapshot()` timing
- `_prepare_snapshot_backtest_input()` has explicit supported/unsupported behavior for binary, directional regression, and multiclass inputs
- docs describe the exact production contract without ambiguity
- tests lock the above in on the canonical suite path

## Related
- #456 umbrella tracking issue
- #460 folded into this issue
- #451 folded into this issue
- #212 older backtest-return correctness background

mikkokotila:
## Summary
- align confusion mean-return metrics to the same immediate-next-execution-row contract used by snapshot backtests
- make post-run snapshot preparation raise explicitly on unsupported multiclass logged outputs
- add focused docs and canonical test coverage for the remaining log/results semantics contract

Closes #463.


limen/log/_permutation_confusion_metrics.py
```diff
@@ -8,30 +8,46 @@
...
+        raise ValueError('execution_lag_bars must be >= 0')
+
+    pred = pd.to_numeric(pd.Series(np.asarray(pred)), errors='coerce')
+    actual = pd.to_numeric(pd.Series(np.asarray(actual)), errors='coerce')
+    open_px = pd.to_numeric(pd.Series(np.asarray(open_px)), errors='coerce')
+    price_change = pd.to_numeric(pd.Series(np.asarray(price_change)), errors='coerce')
+
+    if execution_lag_bars:
+        open_px = open_px.shift(-execution_lag_bars)
+        price_change = price_change.shift(-execution_lag_bars)
```
zero-bang, line 38: Non-blocking observation for future reference: this positional shift assumes the input rows are consecutive bars. That holds on the inline path (`base.py` passes unfiltered price data) and on the post-run path today (default `outlier_quantiles=(0.01, 0.99)` rarely removes rows from small test datasets).

But `_permutation_confusion_metrics` applies outlier filtering to `df` *before* passing `df['open']` / `df['price_change']` here. If filtering removes intermediate rows, `shift(-1)` picks up the next *surviving* row rather than the true next bar, which would break the alignment with the inline path.

For the current default quantiles this is negligible in practice. If you ever tighten the quantiles or use this on sparse data, the fix would be to precompute the shifted price columns on the unfiltered frame before outlier filtering, then join them back.

zero-bang (approved):
Clean, well-scoped PR. The execution-aligned shift in `_confusion_mean_return_pct`, the explicit multiclass rejection in `_prepare_snapshot_backtest_input`, and the test coverage all look correct. I traced all callers of both changed functions — `base.py:115` (inline) and `_permutation_confusion_metrics:188` (post-run) both pick up `execution_lag_bars=1` via the new default, which is the intended alignment.

Test math checks out: the 5-row dummy data is carefully constructed so the shifted values (`[10, -10, -5, 5, NaN]`) produce the expected confusion-quadrant returns, and the feature-bar `tp_x_mean=50.0` vs execution-aligned `tp_mean_return_pct=10.0` split proves the shift is actually working.

One non-blocking forward-looking note inline.

limen/log/_experiment_backtest_results.py
```diff
@@ -14,16 +15,50 @@ def _prepare_snapshot_backtest_input(df: pd.DataFrame) -> pd.DataFrame:
...
+    if 'actuals' in result:
+        actual = pd.to_numeric(result['actuals'], errors='coerce')
+        if _is_unsupported_multiclass(actual):
+            raise ValueError(
+                'snapshot backtest does not support multiclass actuals; '
+                'use binary predictions or directional regression scores'
+            )
+
+    if _is_binary(pred):
+        return result
```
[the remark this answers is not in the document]
mikkokotila: I agree with this one and I will include it in this PR. The intent here is not to get more forgiving; it is to keep the snapshot/log contract explicit and fail hard on corrupted logged outputs rather than silently collapsing them to flat positions. So I am treating this as an in-scope correction to preserve the stricter behavior, not as extra defensive handling.

limen/log/_permutation_confusion_metrics.py
mikkokotila: I agree the observation is real, but I am not widening this PR into outlier-filter execution semantics. For this branch I am keeping the scope to the two contracts in #463 only. I moved this narrower post-run filtering/alignment remainder into #465 and tracked it under umbrella #456 so it does not get lost: #465 covers precomputing the execution-aligned return inputs on the unfiltered frame before outlier filtering, which is the right place to fix it.

limen/log/_experiment_backtest_results.py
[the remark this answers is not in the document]
mikkokotila: Fixed in `b424841`: `_prepare_snapshot_backtest_input()` now fails hard on non-numeric logged prediction/actual values instead of coercing them through to flat positions.

limen/log/_permutation_confusion_metrics.py
mikkokotila: Included now in this PR. `b6d6c14` moves `_confusion_mean_return_pct(...)` above the outlier-filter block so post-run confusion-return metrics stay on the same true next-bar contract as inline. I am leaving #466 as the remaining residue under umbrella #456; #465 is now being resolved by this branch rather than deferred.

zero-bang (approved):

PR #480, opened 2026-04-25
fix: return sample-major cohort probabilities

mikkokotila:
## Summary
- switch optional `return_probs=True` output to sample-major orientation
- keep the main/default `Cohort.predict()` contract unchanged
- add narrow regression coverage for multi-member ordering and single-member shape

## Why
The consumer-facing use case is sample-first: for a given bar/sample, inspect what all members said. This keeps the main sensor-compatible path untouched and changes only the optional probability matrix returned when callers explicitly request member-level introspection.

## Proof
- `python3 -m py_compile limen/cohort/cohort.py tests/test_proto_cohort.py tests/run.py`
- isolated numpy repro against `limen/cohort/cohort.py` verifying:
  - multi-member `return_probs=True` now returns `(n_samples, n_members)`
  - single-member `return_probs=True` now returns `(n_samples, 1)`
- Claude Code CLI review focused on intent and minimality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -764,3 +764,8 @@ Note: add all new changelog entries to the bottom of this file.
 - Raise explicit error when `return_probs=True` is requested in fallback mode
 - Add comprehensive Cohort tests for constructor validation, aggregation behavior, passthrough behavior, and output-option contracts
 - Add integration test coverage validating Cohort as a decoder-compatible replacement at inference
+
+## v2.5.1 on 25th of April, 2026
+
+- Return `Cohort.predict(..., return_probs=True)` probability traces in sample-major order `(n_samples, n_members)` so per-bar cohort inspection matches the supported consumer contract
+- Add regression coverage for sample-major probability traces, `permutation_ids` column ordering, and single-member fallback vote conversion on the canonical `tests.run` path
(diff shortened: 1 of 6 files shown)


zero-bang (approved):
LGTM. Clean, minimal change.

**Correctness**: Both transpose paths are correct — `probs[:, None]` converts 1D `(n_samples,)` to `(n_samples, 1)` for single-member, and `probs_matrix.T` converts `(n_members, n_samples)` to `(n_samples, n_members)` for multi-member. `_format_predict_output` passes `probs_for_return` through as-is, so the shape change lands directly on the caller.

**Column ordering is safe**: `set_members` (line 185) reorders members to match `self.permutation_ids`, so the internal iteration order is always pid-aligned. The test's `list(reversed(sensors))` trick correctly validates this — columns in the returned probs match `permutation_ids` regardless of insertion order.

**No downstream breakage**: No non-test code calls `predict(..., return_probs=True)`, so this is safe to land. Docs, docstring, and all test shape assertions are updated consistently.

**Tests are meaningful**: The multi-member test builds expected probs independently by calling each member's predict and column-stacking in `permutation_ids` order, then asserts `np.allclose` against the returned matrix — this verifies both shape and value correctness. The single-member test confirms the `(n_samples, 1)` column shape. The fallback vote test covers the single-member majority_vote short-circuit with regressor threshold.

PR #489, opened 2026-04-28
fix: use trusted publishing for PyPI release

mikkokotila:
## Summary
- Grant the PyPI publish job GitHub OIDC token permission with `id-token: write`.
- Stop passing legacy `PYPI_USERNAME` / `PYPI_PASSWORD` secrets to `pypa/gh-action-pypi-publish`, allowing PyPI Trusted Publishing to be used.

## Evidence
- Failed run: https://github.com/Vaquum/Limen/actions/runs/25045371268
- Failed step reported PyPI now requires Trusted Publishers or API tokens for automated uploads.
- Local validation: `git diff --cached --check` and YAML parse passed before commit.

## Note
PyPI must have a Trusted Publisher configured for `Vaquum/Limen` and workflow filename `pr_publish_pypi.yml`.

### DIFF
--- a/.github/workflows/pr_publish_pypi.yml
+++ b/.github/workflows/pr_publish_pypi.yml
@@ -96,6 +96,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      id-token: write
 
     steps:
       - name: Download distributions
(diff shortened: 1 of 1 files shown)


zero-bang (approved):
LGTM — clean, minimal switch to OIDC Trusted Publishing. The `id-token: write` permission is correctly scoped to only the publish job, and removing the legacy `user`/`password` params lets the action fall through to OIDC automatically. Once merged, the now-unused `PYPI_USERNAME` and `PYPI_PASSWORD` repo secrets can be deleted.

Issue #490, opened 2026-04-28
Add explicit snapshot backtest support for multiclass and multilabel outputs


mikkokotila:
## Context

Snapshot backtests currently support binary `0/1` predictions directly and directional regression scores through `_prepare_snapshot_backtest_input()`.

They do not currently support multiclass or multilabel outputs:

- multiclass logged predictions/actuals are explicitly rejected by `_prepare_snapshot_backtest_input()`
- multilabel outputs have no supported snapshot semantics and generally fail the current numeric single-column contract

Because snapshot backtest is long-only, support for these output types needs an explicit trading interpretation. It must not silently collapse classes or labels into long/flat decisions.

## Repro

Multiclass currently raises:

```python
_prepare_snapshot_backtest_input(
    pd.DataFrame({
        'predictions': [0, 1, 2],
        'actuals': [0, 1, 2],
    })
)
```

Expected current behavior: `ValueError` stating snapshot backtest does not support multiclass outputs.

Multilabel outputs currently have no documented accepted shape or mapping contract for snapshot backtests.

## Objective

Add explicit snapshot backtest support for multiclass and multilabel model outputs.

## Acceptance Criteria

- Multiclass snapshot backtest support has a documented, deterministic mapping from model output to trading signal or per-class backtest rows.
- Multilabel snapshot backtest support has a documented, deterministic mapping from model output to trading signal or per-label backtest rows.
- Existing binary `0/1` snapshot behavior remains unchanged.
- Existing directional regression snapshot behavior remains unchanged.
- Ambiguous multiclass or multilabel inputs raise clear errors instead of being silently coerced.
- Tests cover supported multiclass inputs, supported multilabel inputs, and unsupported ambiguous inputs.
- Backtest and Log docs describe the supported multiclass and multilabel contracts clearly.

PR #491, opened 2026-04-28
fix: enforce snapshot backtest input contracts

mikkokotila:
## Summary
- Remove the snapshot `trades_count_mode` argument and count trades only as consecutive in-market runs.
- Validate direct snapshot predictions as strict binary `0/1` values and reject empty inputs.
- Reject non-numeric direct snapshot price fields while preserving missing price rows as non-tradable gaps.
- Validate `price_change == close - open` when all three fields are present.
- Add direct regression coverage, docs, changelog, and version bump to `2.5.5`.

## Scope
- No fee/slippage, execution-lag, compounding, drawdown, or output-column changes.
- No multiclass or multilabel support added here; that follow-up is tracked separately in #490.

## Local Evidence
- `.venv/bin/ruff check .`
- `.venv/bin/pytest -q tests/test_metrics_and_log_helpers.py`


zero-bang (approved):
LGTM — verified all three production callers and the test suite are safe against the parameter removal and the stricter validation: `rule_based` and `base` always produce binary 0/1 predictions and compute `price_change` directly from `close - open`, and the experiment-backtest path normalizes through `_prepare_snapshot_backtest_input` before reaching `backtest_snapshot`. Clean, well-scoped change.

zero-bang (approved):

PR #492, opened 2026-04-28
fix: honor experiment_dir in standard UEL runs

### LINKED ISSUE #478: experiment_dir parameter in UniversalExperimentLoop is ignored — experiment saved to CWD instead
### What happened?

When passing experiment_dir to UniversalExperimentLoop, the parameter is silently ignored and the experiment is saved to the current working directory instead of the specified path. No error or warning is raised.

**Limen Version**: `2.4.1`

### To Reproduce
```py
import limen

uel = limen.UniversalExperimentLoop(sfd=my_sfd, experiment_dir="output")
uel.run(
    experiment_name='experiment_test',
    n_permutations=5,
    prep_each_round=True,
    random_search=True
)
```

### Expected Behavior
The experiment should be saved inside the output/ directory as specified by the experiment_dir argument.

### Actual Behavior
The experiment is saved to the current working directory. The output/ directory exists but is not used. No error or warning is produced, making this hard to catch.

### When did it start?

2026-04-22 11:28 UTC

### Severity Level

Low - Other minor issues

mikkokotila:
## Summary
- Fix standard-path UniversalExperimentLoop CSV output so explicit experiment_dir is honored
- Preserve legacy cwd/path-stem CSV behavior when experiment_dir is not set
- Document the standard/MSQ filename distinction and add regression coverage

Closes #478

## Proof
- .venv/bin/ruff check .
- .venv/bin/pytest -q tests/test_experiment_core_standard_csv.py
- .venv/bin/python -m compileall -q limen/experiment/experiment_core.py tests/test_experiment_core_standard_csv.py tests/run.py
- git diff --check


zero-bang (approved):
LGTM — clean fix. The `csv_path` hoisting out of the loop is correct (it was loop-invariant), the `Path(experiment_name).name` strip is the right call when redirecting into `experiment_dir`, and the `mkdir(parents=True, exist_ok=True)` is safe. Test covers both the directory placement and the no-duplicate-header rerun case. Docs and changelog are accurate.

Issue #381, opened 2026-02-12
WP-Limen-0012: General Manifest Clean and Prep {STILL ROOM}


mikkokotila:
**Phase 1: API Alignment** (1 day) (P0)
- [x] Add `with_reference_architecture()` to Manifest class
- [x] Add `with_target_label()` to Manifest class
- [x] Rename internal attributes (architecture_function, architecture_params)
- [x] Add tests for both old and new APIs
- [x] Update foundational SFDs as proof of concept
- [x] Remove `LOOP_ENV` for test mode. Manifest will override their data source in test mode to fetch data from files.
- [x] Refactor Manifest to have clear separation of responsibility between ML and Rule based (base manifest, ml, rule_based)

**Phase 2: Calibration Builder** (1.5 days) (P0)
- [x] Implement CalibrationBuilder class in manifest_core.py
- [x] Add `with_calibration()` method to Manifest
- [x] Implement config injection in `run_model()`
- [x] Update tabpfn_binary to accept calibration_config parameter
- [x] Update logreg_binary to accept calibration_config parameter

**Phase 3: Target Label Transform Convention** (1.5 days) (P0)
- [x] Create limen/label/ module structure
- [x] Implement QuantileBinaryTargetLabel class
- [x] Implement ForwardBreakoutTargetLabel class
- [x] Implement ThresholdBinaryTargetLabel class
- [x] Implement NextReturnTargetLabel class
- [x] Add `with_target_class()` method to Manifest
- [x] Add `with_target_function()` method to Manifest
- [x] Implement target execution logic in manifest core
- [x] Add tests for all target classes
- [x] Add integration tests for target execution
- [x] Migrate Foundational SFD to use enhanced target labeling

PR #497, opened 2026-05-04
feat: CalibrationBuilder and limen.calibration submodule

pdey:

## What does this PR change?
- Add `limen.calibration` submodule with sklearn-based probability calibration and vectorised grid threshold optimisation utilities
- Add `CalibrationBuilder` fluent API to `Manifest` for declarative calibration configuration; supports four modes : raw model, calibration-only, threshold-only, or both,  controlled by per-round flags
- Add explicit callable protocols for plug-and-play custom calibrators and threshold optimisers
- Add calibration support to `LogRegBinary` and `TabPFNBinary` reference architectures; results include `optimal_threshold` and `val_score` when calibration is configured
- Remove old calibration utilities from `limen.transforms` (superseded by `limen.calibration`)
- Fix CSV column alignment when rounds return variable result keys; fix graceful shutdown on SIGINT during a round

Implements calibration builder - phase 2 of #381 

## Checklist


tests/run.py
```diff
@@ -851,7 +868,6 @@
     test_xgboost_evaluate_returns_all_metric_types,
```
zero-bang, line 864: `test_tabpfn_train_evaluate_end_to_end` was removed from the runner but the test function was updated for the new `CalibrationConfig` API and still exists in `test_reference_architecture.py`. I verified it passes in this environment (tabpfn is installed).

The test already has `if TabPFNBinary is None: return` which safely no-ops when tabpfn isn't installed, so there's no reason to exclude it. Without it in the runner, the TabPFN ↔ calibration integration path goes untested even when tabpfn IS available.

Fix: re-add the import and entry:
```python
from tests.test_reference_architecture import test_tabpfn_train_evaluate_end_to_end
```
and add `test_tabpfn_train_evaluate_end_to_end,` back into the test list (after `test_random_binary_train_evaluate_end_to_end,`).

zero-bang (changes requested):
Thorough and well-designed calibration layer. The `CalibrationBuilder` fluent API is clean, the protocol-based design for pluggable calibrators/optimizers is the right abstraction, the 2×2 mode grid with `use_calibration`/`use_threshold` flags is well thought out, and the CSV column alignment fix correctly handles variable-width result dicts. Tests cover all four mode combinations, string param resolution, builder validation, and the end-to-end LogReg calibration path.

One concrete issue to fix before merge — see inline comment.

Broader note: the `tabpfn_binary` foundational SFD now always runs calibration (no `use_calibration`/`use_threshold` in its params, and both default to `True`). This narrows its search space compared to the old version which searched over `use_calibration: [True, False]` and multiple threshold metrics. This may be intentional for simplicity — `logreg_binary` demonstrates the full search pattern — but worth confirming.

tests/run.py
zero-bang: This is still not addressed — `test_tabpfn_train_evaluate_end_to_end` is still missing from both the import block and the test list in `tests/run.py`. The test function exists and was updated for the new CalibrationConfig API, it just needs to be re-added to the runner.

zero-bang (changes requested):
Some items from my previous review still need to be addressed. See my replies on the relevant threads.

tests/run.py
pdey: These should be fixed now. Can you verify?

zero-bang (approved):

limen/experiment/manifest_core.py
```diff
@@ -748,6 +862,11 @@ def resolve_model_kwargs(self, round_params: dict[str, Any]) -> dict[str, Any]:
                     'It must be provided in round_params.'
                 )
 
+        if has_var_keyword:
+            for k, v in round_params.items():
+                if not k.startswith('_') and k not in model_kwargs:
+                    model_kwargs[k] = v
```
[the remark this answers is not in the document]
pdey: reference architecture will use only the params that are required by it, so this should not cause any problems.

PR #500, opened 2026-05-06
feat: load SFD from experiment_dir before falling back to import_module

blahmonkey:

## What does this PR change?
- Add self-contained-experiment SFD load path to [`Trainer.__init__`](limen/experiment/trainer/trainer.py); the new `_load_sfd_module` helper prefers a `<sfd_module_name>.py` file inside `experiment_dir` (loaded via `importlib.util.spec_from_file_location` + `module_from_spec`, with no `sys.path` mutation and the SFD module itself not registered in `sys.modules` under its bare name) and falls back to `importlib.import_module` only when no such file exists
- Forward-looking flows that ship the SFD inside the experiment_dir (e.g. `trainer_prep.py`-style bundle preparation for paper-trade deploy on Praxis) now Just Work without operator-side `PYTHONPATH` wiring or copying SFD files into a sys.path-mounted directory; built-in SFDs referenced by fully-qualified package path (`limen.sfd.foundational_sfd.logreg_binary`, etc.) continue to load via the legacy `importlib.import_module` fallback unchanged
- Add path-traversal hardening: `_load_sfd_module` rejects any `sfd_module` name whose dot-separated segments are not valid Python identifiers, blocking `..`, `/`, `\`, leading/trailing dots and empty segments before either branch runs
- Add 3 test groups pinning the contract — [`test_trainer_loads_sfd_from_experiment_dir_without_sys_path_mutation`](tests/test_trainer.py) stages a tmp dir with a custom SFD `.py` + bare module name and asserts `Trainer(...)` succeeds with `sys.path` and `sys.modules` (for that bare name) unchanged on exit; [`test_trainer_falls_back_to_import_module_when_no_local_sfd_file`](tests/test_trainer.py) stages a tmp dir referencing a packaged SFD by fully-qualified path and asserts the legacy import path still works; [`test_trainer_rejects_path_traversal_and_invalid_module_names`](tests/test_trainer.py) parametrizes 11 hostile/malformed names and asserts each raises `ValueError`
- Add [`docs/TechnicalDebt.md`](docs/TechnicalDebt.md) and TD-001 documenting the residual `import_module` fallback hardening (any module on `sys.path` is namable; allowlist needed before live-trade deploy where the upstream bundle pipeline is not under the same trust boundary as the deploy operator)
- Update [`docs/Trainer.md`](docs/Trainer.md) with a new "SFD Module Resolution" section describing the two-stage load and the validator
- Bump `vaquum_limen` 3.0.2 → 3.0.3

Motivation: Praxis paper-trade deploy crashlooped at sensor-wiring with `ModuleNotFoundError: No module named 'BtcLogRegEVSFD__r0001'` because the SFD `.py` produced by `trainer_prep.py` lives inside the experiment_dir but is referenced by a bare module name in `metadata.json`, and the launcher only puts `STRATEGIES_BASE_PATH` on `sys.path` — not the experiment directories. Operator-side workarounds (extending `PYTHONPATH` per experiment, copying files into `manifests/`, symlinks) all encode the bad pattern into deploy config; this PR fixes the leaky contract upstream so `Trainer(experiment_dir)` is hermetic by design and scales to N experiments without per-experiment env wiring.

## Checklist


zero-bang (approved):
Clean, well-scoped PR. The two-stage SFD load logic is correct, the path-traversal validation is solid (`isidentifier`-per-segment catches all the nasty inputs), and the test coverage pins the right contracts.

One forward-looking observation (non-blocking): `_resolve_model_class` re-imports via `importlib.import_module(architecture_function.__module__)`. For the current Praxis use case this works because the architecture function is imported from an installed package (e.g. `limen.sfd.reference_architecture.logreg_binary`), so `__module__` resolves against `sys.path`. However, if a future self-contained SFD defines its architecture function *inline* (not imported from an installed package), `__module__` would be the bare SFD name which isn't in `sys.modules`, and `train()` → `_resolve_model_class()` would raise `ModuleNotFoundError`. This doesn't need to be fixed now — the hermetic-init contract this PR establishes is the right first step — but it's worth keeping in mind when the scope expands to fully-inline SFDs.

Everything else checks out: changelog, version bump, docs update, TD-001 documentation, and security model are all consistent with the implementation.

PR #507, opened 2026-05-07
fix: limit historical data to latest rows

### LINKED ISSUE #498: fix: make n_rows grab the latest and not the oldest
Right now, n_rows takes the oldest rows. It must take the latest rows.

Let's also rename `n_rows` to `row_count_limit`. 

Let's also add `end_date_limit` which works exactly like `start_date_limit` but for defining the end date. Let's also make it so that if `end_date_limit` and `start_date_limit` are both not None, then `n_rows` must be None.

mikkokotila:
Fixes #498.

## Summary
- Adds `row_count_limit` as canonical row limiter for historical data APIs; keeps `n_rows` as legacy alias.
- Applies row limiting to latest rows after normalization, filtering, and aggregation.
- Adds inclusive `end_date_limit`; date-only end limits include the full named day.
- Rejects row limiting when both date limits define a closed window.
- Updates tests, docs, foundational SFD configs, changelog, and version.

## Proof Coverage
- Latest-row limiting covered for `get_any_file`, `get_spot_klines`, and legacy `n_rows` alias paths.
- `end_date_limit` coverage includes timestamp literals, invalid literals, and date-only full-day semantics.
- Closed-window plus row-limit rejection covered.

## Validation
- `/Library/Frameworks/Python.framework/Versions/3.10/bin/ruff check limen/data/historical_data.py tests/test_historical_data.py` passed.
- `.venv/bin/python -m pytest tests/test_historical_data.py` passed: 10 passed.
- `.venv/bin/python tests/run.py` passed: 584 passed, 0 failed, 45m 51.712s.
- GitHub checks passed on commit `398bb5d`: Ruff, CodeQL, PR validation tests/runtime/coverage, coverage comment, Worker build.


limen/data/historical_data.py
```diff
@@ -491,9 +520,17 @@ def get_spot_klines(
                 pl.col('datetime') >= pl.lit(start_datetime)
             )
 
+        if end_date_limit is not None:
+            end_datetime = datetime.strptime(
+                end_date_limit, '%Y-%m-%d %H:%M:%S'
+            ).replace(tzinfo=timezone.utc)
+            base_data = base_data.filter(
+                pl.col('datetime') <= pl.lit(end_datetime)
```
bit-mis, line 539: P2: `end_date_limit` is asymmetric with `start_date_limit` for date-only literals. `_normalize_datetime_literal` parses `'2020-01-02'` to `'2020-01-02 00:00:00'`, so the filter `<= 2020-01-02 00:00:00 UTC` admits only the midnight tick of Jan 2 — virtually all of Jan 2 is excluded. By contrast, `start_date_limit='2020-01-01'` correctly admits the full target day because midnight aligns with the start of the day for `>=`.

A caller using `start_date_limit='2020-01-01', end_date_limit='2020-01-02'` will reasonably expect ~48h of data and instead receive ~24h plus one tick. Consider rolling a date-only `end_date_limit` forward to end-of-day, or documenting this asymmetry in `docs/Historical-Data.md` next to the existing rule for these limits.

bit-mis (changes requested):

limen/data/historical_data.py
zero-bang: +1 — the root cause is that `_normalize_datetime_literal` applies the same normalization for both limits (`'2020-01-02'` → `'2020-01-02 00:00:00'`), but `>=` naturally aligns with day-start while `<=` does not align with day-end. Rolling a date-only `end_date_limit` to `23:59:59` (or just under next-day midnight) inside `get_spot_klines` before the filter is the cleanest fix — it keeps `_normalize_datetime_literal` generic and only adjusts the comparison endpoint. Alternatively, using `<` with a +1 day offset for date-only inputs would give a true exclusive-end semantic that avoids the off-by-one tick.

zero-bang (commented):
Thorough rename, correct `.head()` → `.tail()` fix, clean legacy alias resolution, and solid test coverage (100% on new statements). The only finding worth addressing is the `end_date_limit` day-boundary asymmetry that bit-mis flagged — I've replied on that thread with implementation suggestions. Everything else checks out: all `n_rows` callers in foundational SFDs, docs, and tests are updated; the `_read_any_file` internal API correctly sheds its `n_rows` param; the closed-window rejection logic is correct; and the signature changes preserve backward compatibility for both positional and keyword callers.

zero-bang (approved):

bit-mis (approved):

PR #511, opened 2026-05-13
feat: add rust ridge benchmark core

mikkokotila:
## Summary
- add a dependency-free Rust ridge core with binary input, sklearn-equivalent intercept handling, and working-set accounting
- add a Limen real-data benchmark script using 15m get_spot_klines, feature prep, sklearn coefficient checks, and 10..100000 training counts
- ignore Rust target build output

## Proof
- cargo fmt --manifest-path rust/ridge_core/Cargo.toml --check
- cargo test --manifest-path rust/ridge_core/Cargo.toml
- cargo build --release --manifest-path rust/ridge_core/Cargo.toml
- .venv/bin/python -m ruff check scripts/bench_ridge_core.py
- .venv/bin/python scripts/bench_ridge_core.py

## Benchmark Evidence
- data: 2025-01-01..2025-02-01 BTCUSDT spot 15m, 2,975 prepared rows, 12 features
- sklearn coefficient max abs diff: 4.828312150095826e-08 across alphas 1e-6, 1e-3, 1, 100
- working set: 309,400 input bytes; 312,312 full training live bytes
- 100000 full trainings: 21.173439333s total, 211.734393 us/train

## Review
- Codex 5.5 xhigh: approved after blocker fixes
- Claude Opus Max: approved pre-production benchmark reporting


rust/ridge_core/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    #[test]
+    fn binary_roundtrip() {
+        let x = vec![1.0, 2.0, 3.0, 4.0, 5.0, 6.0];
+        let y = vec![1.0, 2.0, 3.0];
+        let bytes = write_binary(3, 2, &x, &y).unwrap();
+        let problem = RidgeProblem::from_binary(&bytes).unwrap();
+        assert_eq!(problem.rows(), 3);
+        assert_eq!(problem.cols(), 2);
```
bit-mis, line 514: **p2 — test does not exercise data preservation.** This roundtrip only asserts dimensions, but `rows`/`cols` are read from the 24-byte header (separate from the f64 payload). If `write_binary` or `from_binary` ever silently corrupted `x` or `y` bytes — wrong endianness, off-by-one in the offset advance, mis-stride on multi-column rows, etc. — this test would still pass green. Suggest training a model from the roundtripped problem and asserting `beta` matches one trained directly from the original `(x, y)` arrays so the assertion actually covers the payload path. The Python integration covers this against sklearn, but the unit test as written gives false confidence in isolation.

bit-mis (changes requested):

bit-mis (approved):

PR #512, opened 2026-05-15
feat: add canonical decoder outcome targets

mikkokotila:
## Summary
- add canonical decoder outcome targets: `NextBarUpTarget`, `NextBarDownTarget`, and `VolNormalizedReturnTarget`
- export the targets through `limen.targets` and document manifest usage
- bump package version to `3.1.1` and update `CHANGELOG.md`

## Target Semantics
- `next_bar_up`: `1` when next close is above current close, else `0`; terminal row is null
- `next_bar_down`: `1` when next close is below current close, else `0`; terminal row is null
- `vol_normalized_return`: log return divided by prior-bar Parkinson sigma, with dirty OHLC rows dropped and zero sigma nulled

## Proof
- `git diff --cached --check`
- `.venv/bin/ruff check limen/targets tests/test_targets.py tests/run.py`
- `.venv/bin/python -m pytest tests/test_targets.py` — 28 passed
- `.venv/bin/python tests/run.py` — 591 passed, 0 failed
- GitHub PR checks — pass: Tests, Runtime, Coverage, Ruff, CodeQL, Workers, Publish Coverage Comment

## Notes
- Bucketing remains downstream-only: fit train `q33`/`q66` of `abs(z)`, persist boundaries, and never refit online.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -879,3 +879,10 @@ Note: add all new changelog entries to the bottom of this file.
 - Update `docs/Features.md` to drop the moved `ema_breakout` row and `docs/Targets.md` to document `EmaBreakoutTarget`, `ExitQualityTarget`, `RiskRewardRatioTarget`
 - Raise `tests/runtime_budget.json` ceiling from 180s to 195s to accommodate the small per-test overhead introduced by the added `when/then/EPSILON` guards across the indicator/feature surface
 - `ExitQualityTarget` and `RiskRewardRatioTarget` drop their forward-looking source columns (`exit_reason`, `exit_net_return` and `capturable_breakout`, `max_drawdown` respectively) after computing the target so they cannot re-enter `x_train`/`x_val`/`x_test` via the manifest's all-columns-except-target convention
+
+## v3.1.1 on 17th of May, 2026
+
+- Add canonical decoder outcome targets `NextBarUpTarget`, `NextBarDownTarget`, and `VolNormalizedReturnTarget`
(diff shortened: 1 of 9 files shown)


zero-bang (approved):
LGTM — clean, correct implementation that follows the established target interface contract.

**What I verified:**

- **NextBarUpTarget / NextBarDownTarget**: Simple shift-and-compare logic is correct. Terminal null from `shift(-1)` is handled naturally by Polars cast semantics. Flat bars (equal close) correctly produce 0 for both targets (neither "above" nor "below"), which matches the documented semantics.

- **VolNormalizedReturnTarget math**: Parkinson scale constant `4·ln(2)` is correct. Log-return computation, EWMA sigma, one-bar sigma shift to prevent current-bar leakage, and zero-sigma → null conversion are all implemented correctly. The sanity gate median ratio check is reasonable.

- **Dirty bar filter**: `high >= max(open, close) AND low <= min(open, close)` is the correct OHLC consistency check. Row-dropping in `transform` is safe because it operates on the full DataFrame (features + target together), so the manifest sees a consistent slice.

- **Train-only fitting**: Constructor validates on `train_data`; the fitted instance is reused for val/test via `_apply_class_based_target`. Consistent with existing targets like `QuantileBinaryTarget`.

- **Tests**: Traced the `_constant_vol_bars` helper math — Parkinson variance collapses to `log_return²` when `range_scale=1.0`, giving unit normalized return. Prior-sigma-shift test correctly validates that mutating row 2's high changes row 3 but not row 2. Dirty bar test, rejection test, and manifest integration test all exercise the right code paths.

- **Docs, exports, changelog, version bump**: All consistent and correctly ordered.

PR #514, opened 2026-05-17
fix: v3.1.0 — leak fixes and target migration for indicators/features

mikkokotila:
## Summary

Addresses one critical look-ahead leak in `ichimoku_cloud`, migrates three forward-looking helpers from `limen.features` to `limen.targets` as class-based targets, and applies twelve correctness fixes for issues surfaced in a full audit of the indicator/feature surface (~205 files reviewed).

## Critical

- **`ichimoku_cloud.py`** — `senkou_a`/`senkou_b` were written with `.shift(-displacement)`, reading data from `t+displacement` (future) at row `t`. Changed to `.shift(displacement)` so the value at row `t` is computed from `t-displacement` (past). Streaming-safe. Test updated to the corrected expected values.
- **3 file moves** — `ema_breakout`, `exit_quality`, `risk_reward_ratio` were forward-looking labels living in the features namespace where a "select all feature columns" pipeline could silently train on the answer. Migrated to `limen.targets.{EmaBreakoutTarget, ExitQualityTarget, RiskRewardRatioTarget}` as classes matching the existing `ForwardBreakoutTarget` pattern (`__init__(train_data, target_name)`, `transform(data, **kwargs)`).

## High

- `wilder_rsi` reseeded via canonical `rsi()` (canonical Wilder smoothing; deviated previously)
- `bollinger_bands` uses `ddof=0` to match `bbands` TA-Lib parity
- `stochastic_oscillator` adds flat-window zero guard + preserves warm-up nulls (matches `willr.py` pattern)
- `conserved_flux_renormalization` guards `flux_mean == 0.0 or is None`
- `breakout_features` removes silent `except (ValueError, IndexError)` fallback
- `quantile_line_density` output column parameterized as `quantile_line_density_{lookback_hours}h`
- `active_quantile_count` half-open interval unified to `end_idx + 1` (matches `active_lines`)
- `hours_since_big_move` boundary unified to `<=` (matches `hours_since_quantile_line`)
- `ht_trendline` / `ht_trendmode` add `if idx < 0: break` to inner accumulator (prevents numpy negative-index wraparound)
- EPSILON denominator guards added to `range_pct`, `spread`, `spread_percent`, `sma_ratios`, `trend_strength`, `volume_ratio`, `volume_regime`, `volume_spike`, `volume_trend`, `volume_weight`, `vwap`, `close_to_extremes` features and `body_pct` indicator

## Test plan

- [x] All 219 targeted tests pass (`test_feature_library_*`, `test_indicator_*`, `test_indicators_vs_talib`, `test_targets`, `test_conserved_flux_renormalization`)
- [x] All 31 `test_trainer.py` tests pass (exercises every foundational SFD that consumes the changed indicators: `tabpfn_binary` uses `bollinger_bands`, `logreg_binary`/`tabpfn_binary`/`rule_based` use `wilder_rsi`, `xgboost_regressor` uses `stochastic_oscillator`)
- [x] All 19 `test_scalers.py` tests pass
- [x] `ruff check limen/` clean
- [x] Affected tests updated to new expected values: `test_ichimoku_cloud_builds_shifted_components` (new leak-free senkou values), `test_active_quantile_count_tracks_quantile_filtered_spans` (`+1` boundary), `test_quantile_line_density_counts_recent_line_endings_within_lookback` (parameterized column name)
- [x] Three former-feature tests moved to `tests/test_targets.py` against the new target classes
- [x] Self-review pass with `codex review --base main` and `claude --print` — both caught the original stochastic_oscillator warm-up null regression; fix applied and re-verified


zero-bang (approved):
Thorough, well-structured PR. The ichimoku look-ahead fix is correct, the feature-to-target migrations preserve semantics, and the EPSILON / boundary / warm-up guards are all sound. I verified callers, scalers, and test expectations across the affected surface.

One non-blocking observation on `ht_trendline` / `ht_trendmode`: the new `if idx < 0: break` guard prevents numpy negative-index wraparound (clearly correct), but the subsequent `temp_real / float(dc_period_int)` still divides by the intended period rather than the actual accumulated count. In theory this would bias the average low if the guard fires. In practice, the 63-period lookback and slow `smooth_period` ramp-up make this unreachable for reasonable data — `dc_period_int` stays well below `today` throughout the main loop. The guard is strictly an improvement over the status quo. Worth a follow-up if you ever relax the lookback.

LGTM — approving.

mikkokotila:
@zero-bang Pushed a small follow-up commit ([872c7bc](https://github.com/Vaquum/Limen/pull/514/commits/872c7bc)) to address Copilot's inline note on `wilder_rsi.py:L21`: the previous `rsi(...).rename(...)` form would have clobbered any pre-existing `rsi_{period}` column when both indicators ran on the same DataFrame. New implementation calls `_rsi_from_values` directly with the `wilder_rsi_{period}` alias, leaving any existing `rsi_{period}` column untouched. Verified locally: both columns coexist correctly when stacked, and values match for the same period. Could you take a quick look?

zero-bang:
Reviewed the follow-up commit (872c7bc). The fix is correct — calling `_rsi_from_values` directly and aliasing to `wilder_rsi_{period}` avoids the `rsi(...).rename(...)` intermediate that would have clobbered any pre-existing `rsi_{period}` column. Both columns now coexist cleanly when stacked on the same DataFrame. LGTM.

mikkokotila:
Pushed [e05ce82](https://github.com/Vaquum/Limen/pull/514/commits/e05ce82): bumped `tests/runtime_budget.json` from 180s to 200s. The cumulative polars `when/then`/EPSILON guards introduced in this PR push the 584-test suite from ~178s to 181.4s on the CI runner — over the 180s ceiling by 1.4s. Re-running confirmed the same recorded time, so it's a real (small) suite-cost increase, not flake. 200s restores comfortable headroom (~18s) consistent with the precedent in [86061fa](https://github.com/Vaquum/Limen/commit/86061fa) (the previous 135→180 bump). @zero-bang FYI.

zero-bang:
Acknowledged. The 180→200s bump is justified — 1.4s of real overhead from the cumulative guards, and the 200s ceiling keeps the same ~10% headroom ratio as the prior 135→180 bump. No concerns.

mikkokotila:
Pushed [0d7a63f](https://github.com/Vaquum/Limen/pull/514/commits/0d7a63f) addressing Copilot's two inline findings on the outcome-target classes: `ExitQualityTarget` and `RiskRewardRatioTarget` were leaving their forward-looking source columns (`exit_reason`/`exit_net_return` and `capturable_breakout`/`max_drawdown`) in the returned frame. The manifest's "every column except target_name" preparation step would have re-introduced those into `x_train`/`x_val`/`x_test`, defeating the whole point of moving these helpers to targets. Both transforms now drop their source columns after the target is computed. Tests extended to assert the columns are absent from the result.

Note: `EmaBreakoutTarget` consumes `target_col` which is a legitimate feature column (`close` by default), so it's left in place — same as how `ForwardBreakoutTarget` handles `close`. @zero-bang FYI.

zero-bang:
Copilot flagged two valid points on `RiskRewardRatioTarget` and `ExitQualityTarget`: the source columns (`capturable_breakout`, `max_drawdown`, `exit_reason`, `exit_net_return`) remain in the DataFrame after `transform()`, and since `split_data_to_prep_output` at `limen/data/utils/splits.py:105` slices `cols[:-1]` as features, those columns end up in `x_train`/`x_val`/`x_test`.

This is a real gap, though worth noting the context:

1. **This PR is still a net improvement.** Before, the *derived* columns (`risk_reward_ratio`, `exit_quality`) lived in `limen.features`, so both the source columns *and* the computed label would appear as features. Now the label is excluded by the target mechanism — the leak surface is smaller.

2. **The source columns are pre-existing in the user's data** (post-trade fields from backtesting output), so they'd leak as features regardless of whether the target drops them. The target transform doesn't introduce them — it just consumes them.

3. **No existing target drops source columns.** `ForwardBreakoutTarget` and `NextReturnTarget` use `close`, which happens to be a legitimate feature, so the pattern hasn't been tested against non-feature source columns.

That said, the right fix is straightforward: either (a) drop the source columns in `transform()`, or (b) add a `source_columns` attribute to the target interface and have the manifest drop them in `_apply_class_based_target()`. Option (b) is cleaner since it keeps the convention explicit and works for any future target that consumes non-feature columns.

This seems like a reasonable follow-up issue rather than a blocker for this PR — the current state is strictly better than before, and the architectural fix (source-column exclusion mechanism) should be designed once rather than ad-hoc'd per target. Happy to pick up that follow-up if useful.

mikkokotila:
@zero-bang Looks like a crossed-wire timing thing — I pushed [0d7a63f](https://github.com/Vaquum/Limen/pull/514/commits/0d7a63f) at 09:22 UTC, just before your comment at 09:24 UTC, and it already implements your option (a): both `ExitQualityTarget.transform()` and `RiskRewardRatioTarget.transform()` now `.drop()` the source columns after creating the target. Tests assert the source columns are absent from the result.

Your option (b) — adding a `source_columns` attribute to the target interface so the manifest handles exclusion once in `_apply_class_based_target()` — is the cleaner architectural fix and I agree it should be designed once rather than ad-hoc'd per target. Want me to file that as a follow-up issue rather than fold it into this PR? Your three points hold either way: this PR is a net improvement, the source columns existed in user input independently, and the existing targets haven't been tested against non-feature source columns.

For now, all CI is green on `0d7a63f` (Tests, Runtime, Coverage, CodeQL, Ruff, Workers all pass).

zero-bang:
Ha — yes, crossed wires. I see `0d7a63f` now; the `.drop()` calls in both transforms are exactly what I was about to suggest. Tests asserting column absence seal it. No further concerns on this PR.

Agreed on filing option (b) as a follow-up issue. The per-target `.drop()` is correct today, but a `source_columns` attribute on the interface would make the contract explicit and let the manifest enforce it once in `_apply_class_based_target()` — worth designing properly rather than rushing into this already-large PR. Happy to review that when it lands.

mikkokotila:
Filed [#515](https://github.com/Vaquum/Limen/issues/515) for the option (b) follow-up (`source_columns` attribute on the target interface + manifest-level enforcement). All checks green on `0d7a63f`; the formal APPROVE state is still on the earlier `c58834e` since each follow-up commit invalidated it. @zero-bang would you mind clicking APPROVE on the latest commit when you have a moment? Then this can merge.

zero-bang (approved):
Re-approving on the latest commit (0d7a63f). Reviewed the two follow-up commits (872c7bc, e05ce82) and the source-column drop fix (0d7a63f) — all correct. All 8 CI checks green. LGTM, ready to merge.

zero-bang:
Done — re-approved on `0d7a63f`. All 8 checks green (Tests, Runtime, Coverage, CodeQL, Ruff, Workers, Publish Coverage Comment). Ready to merge whenever you like. Follow-up tracked in #515.

PR #519, opened 2026-05-17
feat: add manifest pca compression

mikkokotila:
## Summary
- add optional MLManifest.set_pca_compression() after scaling and split-column alignment
- replace x_train/x_val/x_test with pc_* columns only when auto_pca is true
- require fitted RobustScaler, train-only PCA fit, frozen validation/test projection
- split PCA scaler diagnostics for missing scaler, scaler-param mismatch, and wrong scaler type
- update docs, changelog, and version to 3.2.1

## Proof
- .venv/bin/ruff check limen/experiment/manifest_core.py tests/test_manifest_prepare_data.py tests/run.py
- git ls-files "*.py" | xargs .venv/bin/ruff check
- .venv/bin/python -m pytest tests/test_manifest_prepare_data.py -k pca_compression -q
- .venv/bin/python -m pytest tests/test_manifest_prepare_data.py tests/test_scalers.py -q
- GitHub PR checks: Ruff, Tests, Runtime, Coverage, CodeQL, Workers all passing

## Notes
- no parallel raw-feature dataset is retained
- .claude/ and rust/ are pre-existing untracked local files and are not part of this PR

### DIFF
--- a/limen/experiment/manifest_core.py
+++ b/limen/experiment/manifest_core.py
@@ -15,12 +15,15 @@ from limen.calibration.pipeline import ThresholdOptimizerProtocol
 if TYPE_CHECKING:
     from limen.sfd.rule_based.config import RuleBasedConfig
 
+import numpy as np
 import polars as pl
+from sklearn.decomposition import PCA
 
 from limen.data.utils import split_by_dates
 from limen.data.utils import split_data_to_prep_output
 from limen.data.utils import split_data_to_rule_based_prep_output
 from limen.data.utils import split_sequential
+from limen.scalers.robust_scaler import RobustScaler
 from limen.scalers.registry import SCALER_REGISTRY
 logger = logging.getLogger(__name__)
 
@@ -48,6 +51,18 @@ class AblationConfig:
 
     drop_count_key: str
     seed_key: str
+
+
+@dataclass
+class PCACompressionConfig:
+
+    '''Configuration for optional manifest-level PCA feature compression.'''
+
+    enabled_param: str
+    n_components_param: str
+    scaler_param_name: str
+    component_prefix: str
+
 
 FittedTransformEntry = tuple[
     list[FittedParamsComputationEntry],
@@ -759,6 +774,7 @@ class MLManifest(Manifest):
 
     scaler: FittedTransformEntry = None
     ablation_config: AblationConfig | None = None
+    pca_compression_config: PCACompressionConfig | None = None
     data_dict_extension: Callable = None
     prediction_calibration_config: CalibrationConfig | None = None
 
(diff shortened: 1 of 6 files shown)


limen/experiment/manifest_core.py
```diff
@@ -1374,6 +1430,126 @@ def _apply_scaler(
...
+        return split_data, all_fitted_params
+
+    if config.n_components_param not in round_params:
+        raise ValueError(
+            f"round_params['{config.n_components_param}'] is required when "
+            f"round_params['{config.enabled_param}'] is True"
+        )
+
+    scaler = all_fitted_params.get(config.scaler_param_name)
+    if not isinstance(scaler, RobustScaler):
```
bit-mis, line 1465: P2 — this single check collapses three distinct failure modes into one misleading error:

1. `set_scaler` was never called → `scaler is None`
2. `set_scaler(RobustScaler, param_name='custom')` was used, leaving `_scaler` unset → `scaler is None` again (the user did configure RobustScaler, just under a different key)
3. A non-RobustScaler was fitted (e.g. `LinearScaler`)

All three currently produce the same message that asserts "requires a fitted RobustScaler" and suggests `scaler_type="robust"`. That suggestion only works with `set_scaler_from_params`, not the `set_scaler` path that both the docstring of `set_pca_compression` and the new docs page recommend. Case (2) is the most painful: the user reads the error, sees they already configured `RobustScaler`, and has no hint that the `scaler_param_name` they passed (or defaulted to) does not match the `param_name` on their `set_scaler` call.

Splitting the check makes the failure self-diagnosing without adding meaningful complexity:

```python
scaler = all_fitted_params.get(config.scaler_param_name)
if scaler is None:
    raise ValueError(
        f"PCA compression could not find a fitted scaler at "
        f"all_fitted_params['{config.scaler_param_name}']. "
        f"Configure .set_scaler(RobustScaler) (or pass a matching "
        f"scaler_param_name to .set_pca_compression)."
    )
if not isinstance(scaler, RobustScaler):
    raise ValueError(
        f"PCA compression requires a fitted RobustScaler at "
        f"all_fitted_params['{config.scaler_param_name}'], got "
        f"{type(scaler).__name__}."
    )
```

Neither case (1) nor case (2) is covered by the new tests in `tests/test_manifest_prepare_data.py` — only `test_pca_compression_requires_robust_scaler` (case 3) is exercised — so this is also a test-coverage gap on top of the message issue.

bit-mis (changes requested):

zero-bang (approved):
Thorough review of the PCA compression implementation. Clean work.

**Pipeline placement** — correct: PCA runs after `_align_split_columns` (all splits have identical column structure) and before `_finalize_to_data_dict`. Train-only fit with frozen projection for val/test is properly implemented.

**Validation** — comprehensive: bool-masquerading-as-int guard (`isinstance(k, bool)`), k bounds against both feature count and train row count, RobustScaler type gate, non-numeric column detection, empty-split handling via `np.empty((0, k))`.

**Integration** — no downstream breakage: `_build_pca_split` preserves `datetime` and target column in the correct order, so `_finalize_to_data_dict` → `split_data_to_prep_output` works unchanged. `price_data_for_backtest` is unaffected since PCA doesn't alter datetimes. The `all_fitted_params` merge into `data_dict` carries `_pca*` metadata cleanly.

**Tests** — 7 tests covering the success path (enabled, disabled, empty splits, train-only fit) and error paths (wrong scaler, invalid k, non-numeric features). The train-only-fit test is well-designed — mutating `close` at index ≥ 24 targets only val/test rows under a (3,1,1) split of 40 rows, confirming PCA components are identical.

**Docs** — accurate and consistent with the implementation.

LGTM.

limen/experiment/manifest_core.py
```diff
@@ -15,12 +15,15 @@
...
+import numpy as np
 import polars as pl
+from sklearn.decomposition import PCA
 
 from limen.data.utils import split_by_dates
 from limen.data.utils import split_data_to_prep_output
 from limen.data.utils import split_data_to_rule_based_prep_output
 from limen.data.utils import split_sequential
+from limen.scalers.robust_scaler import RobustScaler
 from limen.scalers.registry import SCALER_REGISTRY
```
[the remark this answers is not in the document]
mikkokotila: Addressed in d7cd188: moved local calibration imports below third-party imports while keeping TYPE_CHECKING adjacent to typing.

limen/experiment/manifest_core.py
mikkokotila: Addressed in d7cd188: split missing scaler from wrong scaler diagnostics and added tests for no scaler and scaler_param_name mismatch.

bit-mis (approved):

zero-bang (approved):

PR #522, opened 2026-05-18
chore: bump version to 3.3.0

mikkokotila:
## What does this PR change?

Updates release metadata for PR #520 to **3.3.0**:

- Bumps `pyproject.toml` from `3.2.1` to `3.3.0`
- Adds the missing `CHANGELOG.md` entry for the snapshot backtest metric-ledger change

This addresses the review finding that the output-schema change needs a minor bump, not a patch bump.

## Validation

- Reviewed the full diff
- No local test suite run per operator direction
- Dry compile only: `python3 -m compileall -q limen tests`
- Metadata parse returned `3.3.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "vaquum_limen"
-version = "3.2.1"
+version = "3.3.0"
 description = "Bitcoin-first research and trading platform."
 readme = "README.md"
 authors = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "vaquum_limen"
-version = "3.2.1"
+version = "3.2.2"
```
bit-mis: **P2 — version bump understates the documented change**

This is a PATCH bump (`3.2.1` → `3.2.2`), but the CHANGELOG entry below documents non-backward-compatible behavior from PR #520:

- Snapshot backtest output is *replaced* with the 22-column ledger (consumers expecting the prior columns break).
- `rule_based_metrics` renames `drawdown_std` → `drawdown_std_bps` and drops `sharpe_std` / `sharpe_degradation` from its output dict; `is_stable` is hard-coded to `False`.

Per SemVer (which CLAUDE.md mandates for this CHANGELOG), backward-incompatible output schema changes warrant at minimum a MINOR bump (`3.3.0`), not PATCH. Recommend bumping to `3.3.0` here so downstream pinning (e.g. `vaquum_limen~=3.2`) does not silently absorb the schema break.

bit-mis (changes requested):

pyproject.toml
mikkokotila: Addressed. The PR now bumps `pyproject.toml` and `CHANGELOG.md` to `3.3.0`, and the PR title/body state the minor-version target explicitly.

bit-mis (approved):

Issue #506, opened 2026-05-07
WP-Limen-0013: YAML based Experiment


pdey:
Implements RFC-1005. Adds a declarative YAML experiment format, a compiler that converts YAML to a runnable SFD, and a CLI (`limen run/validate/profile/init`). Provides the API foundation for the future GUI experiment builder.

**Depends on:** RFC-1004 complete (CalibrationBuilder, MLManifest/RuleBasedManifest split — merged to main)

**NOTE:** RFC-1005 was written before several Manifest features were added. This issue extends the YAML schema to cover the full MLManifest surface (`set_pre_split_data_selector`, `set_bar_formation`, `set_feature_ablation`, `add_to_data_dict`, `set_scaler_from_params`, `with_params_override`, `metrics_params`) and adds full `RuleBasedManifest` support via a `type` field. `limen.calibration` is added to the resolver's `ALLOWED_NAMESPACES` (missing from the RFC spec).

---

## Non-coding (1 day)

- [ ] Finalize complexity rating thresholds (permutation bounds for low / medium / high / extreme)
- [ ] Confirm YAML schema v1.0 field list — cross-check against full MLManifest and RuleBasedManifest surface

---

## Phase 1: Core Module (3 days)

- [x] **PR 1.1: Scaffold + Errors + Schema** (1 day)
  - Create `limen/yaml/` module structure and `__init__.py` public API
  - Implement `errors.py` — `YAMLError`, `ValidationError`, `ResolutionError`, `GitError`
  - Implement `schema.py` — `VERSION` constant and full schema dict covering all MLManifest and RuleBasedManifest fields; `type: "ml" | "rule_based"` field determines which manifest subclass is built
  - Add `ruamel.yaml` and `click` to `pyproject.toml` dependencies

- [x] **PR 1.2: Parser + Resolver** (2 days)
  - Implement `parser.py` using `ruamel.yaml` with line number tracking
  - Implement `resolver.py` with namespace validation
  - `ALLOWED_NAMESPACES` includes `limen.calibration` (missing from RFC-1005 spec) and `limen.sfd.rule_based` (for rule-based predicate functions)
  - Unit tests for parser (valid YAML, syntax errors, line tracking)
  - Unit tests for resolver (allowed/blocked namespaces, non-existent paths)

---

## Phase 2: Validation + Git Integration (4 days)

- [x] **PR 2.1: Validator** (2 days)
[...]

PR #523, opened 2026-05-18
Wp limen 0013 yaml experiment

pdey:

## What does this PR change?

- Add `limen/yaml/` module: YAML experiment definition format with parser, resolver, schema, validator, and compiler
- Add `limen/cli/` entry point with `limen validate` and `limen run` CLI commands (Click); `limen run --dry-run` validates and compiles without executing
- YAML experiments are fully declarative: `sfd.manifest` compiles to `MLManifest` or `RuleBasedManifest`; `sfd.params` defines the parameter search space; `uel` configures the `UniversalExperimentLoop`
- `{param_name}` syntax passes round_param references through as strings at compile time; `limen.*` dotted paths are resolved to callables
- Validator collects all errors without fail-fast: required fields, schema version, resolvable `limen.*` paths, param coverage, calibration cross-references
- `split_config` (ratio-based) and `split_dates` (absolute ISO-date bounds) are mutually exclusive alternatives in the manifest; exactly one is required
- `pca_compression` optional dict in the manifest wires `MLManifest.set_pca_compression()` with all four params defaulted
- `logreg_binary.yaml` template added to `limen/yaml/templates/`
- Results directory copies the source YAML for reproducibility; `uel.output_path` supports `{name}` and `{datetime}` placeholders

Implements  core yaml based experiment design of #506 

## Checklist

- [ ] I updated `/docs` (if behavior/API/config/user/etc changed)


limen/experiment/manifest_core.py
```diff
@@ -1157,7 +1157,14 @@ def _resolve_params(params: dict[str, Any], round_params: dict[str, Any]) -> dic
             if value.startswith('_') or value in round_params:
                 resolved[key] = round_params[value]
             elif '{' in value and '}' in value:
-                resolved[key] = value.format(**round_params)
+                import re as _re
+                m = _re.fullmatch(r'\{(\w+)\}', value.strip())
+                if m:
+                    # pure {param_name} reference — direct lookup preserves type (float, int, bool)
```
zero-bang: `import re as _re` should be a module-level import. This sits inside `_resolve_params` which is called per-round per-param during experiment execution. While Python caches imports after the first call, placing `import` inside a loop body is non-standard and makes the dependency invisible at the top of the file.

Move `import re` to the top of `manifest_core.py` with the other imports, and use `re.fullmatch(...)` directly.

limen/yaml/rules.py
```diff
@@ -0,0 +1,600 @@
...
+        if not isinstance(params, dict):
+            errors.append(YAMLError(
+                message="'sfd.params' must be a mapping",
+                path='sfd.params',
+                suggestion='Each key maps to a list of values, e.g. lookback: [12, 24, 48]',
+            ))
+            return
+
+        for key, values in params.items():
+            if not isinstance(values, list):
```
zero-bang: The docstring says "Every value in sfd.params must be a **non-empty** list" but the check only validates `isinstance(values, list)`. An empty list `[]` passes validation here but `ParamDomain.__init__` raises `ValueError` at runtime for empty value lists, producing a confusing error.

Add an emptiness check:
```python
if not isinstance(values, list) or len(values) == 0:
    errors.append(YAMLError(
        message=f"Parameter '{key}' must be a non-empty list of values",
        path=f'sfd.params.{key}',
        suggestion=f'Add at least one value, e.g. {key}: [1]',
    ))
```

limen/yaml/rules.py
```diff
@@ -0,0 +1,600 @@
...
+            return
+
+        for i, cond in enumerate(conditions):
+            if not isinstance(cond, dict):
+                errors.append(YAMLError(
+                    message='Each condition must be a mapping',
+                    path=f'sfd.manifest.strategy.conditions[{i}]',
+                ))
+                continue
+            for key in ('name', 'type', 'expression'):
```
zero-bang: `ConditionsList` requires `('name', 'type', 'expression')` for every condition, but the test fixture `_MINIMAL_RULE_BASED_YAML` uses `type: threshold` conditions with `column/operator/value` fields and no `expression`. No test calls `validate()` on the rule-based fixture, so this mismatch is invisible.

Either:
- The required fields should vary by condition type (e.g. `polars_lambda` needs `expression`, `threshold` needs `column`/`operator`/`value`), or
- The test fixture should include an `expression` field and match the schema

Please also add a `test_validate_passes_valid_rule_based_yaml` test that runs `validate()` on the rule-based fixture to catch this kind of gap.

limen/yaml/schema.py
```diff
@@ -0,0 +1,78 @@
...
+VALID_SEARCH_STRATEGY_TYPES = {'random', 'grid'}
+
+TOP_LEVEL_REQUIRED = {'schema_version', 'metadata', 'sfd', 'uel'}
+
+METADATA_REQUIRED = {'name', 'limen_version', 'mode'}
+METADATA_OPTIONAL = {'author', 'created_at', 'description', 'tags'}
+
+SFD_REQUIRED = {'manifest', 'params'}
+
+MANIFEST_REQUIRED = {'type', 'data_source', 'split_config', 'reference_architecture'}
```
zero-bang: `MANIFEST_REQUIRED` includes `split_config`, but now that `split_dates` is a mutually exclusive alternative (enforced by `SplitSpec`), `split_config` is not always required. Since this constant is used in `NoUnknownKeys` for key recognition, `split_config` should move to `MANIFEST_OPTIONAL_SHARED` alongside `split_dates` to avoid misleading future readers:

```python
MANIFEST_REQUIRED = {'type', 'data_source', 'reference_architecture'}
MANIFEST_OPTIONAL_SHARED = {'test_data_source', 'required_columns', 'split_config', 'split_dates'}
```

limen/cli/commands/run.py
```diff
@@ -0,0 +1,125 @@
...
+    click.echo(f"  Results → {results_dir}")
+
+    uel = UniversalExperimentLoop(
+        sfd=compiled,
+        search_strategy=search_strategy,
+        experiment_dir=results_dir,
+        test_mode=test_mode,
+    )
+
+    uel.run(
```
zero-bang: If `uel.run()` or `CompiledSFD()` raises an exception, the user gets a raw traceback instead of a CLI-friendly error message. Consider wrapping the execution block (lines 67–98) in a try/except that catches `Exception`, prints a clean error via `click.secho`, and returns `False`:

```python
try:
    uel = UniversalExperimentLoop(...)
    uel.run(...)
except Exception as exc:
    click.secho(f'  ✗ Runtime error: {exc}', fg='red')
    return False
```

zero-bang (changes requested):
Solid foundation — the rule engine, compiler, and CLI are well-structured and the test coverage for the ML path is thorough. A few issues to address before merging:

1. **`import re` inside hot-path function** — should be module-level
2. **`SfdParams` doesn't reject empty param lists** — passes validation but crashes `ParamDomain` at runtime
3. **Rule-based validation is untested** — the test fixture conditions lack the `expression` field that `ConditionsList` requires, and no test runs `validate()` on the rule-based YAML
4. **`MANIFEST_REQUIRED` includes `split_config`** — misleading now that `split_dates` is a mutually exclusive alternative
5. **`run_experiment` doesn't catch runtime exceptions** — `uel.run()` errors produce raw tracebacks instead of CLI-friendly output

zero-bang (approved):

CHANGELOG.md
```diff
@@ -897,3 +897,15 @@ Note: add all new changelog entries to the bottom of this file.
...
+- Add `limen/cli/` entry point with `limen validate` and `limen run` CLI commands (Click); `limen run --dry-run` validates and compiles without executing
+- YAML experiments are fully declarative: `sfd.manifest` compiles to `MLManifest` or `RuleBasedManifest`; `sfd.params` defines the parameter search space; `uel` configures the `UniversalExperimentLoop`
+- `{param_name}` syntax passes round_param references through as strings at compile time; `limen.*` dotted paths are resolved to callables
+- Validator collects all errors without fail-fast: required fields, schema version, resolvable `limen.*` paths, param coverage, calibration cross-references
+- `split_config` (ratio-based) and `split_dates` (absolute ISO-date bounds) are mutually exclusive alternatives in the manifest; exactly one is required
+- `pca_compression` optional dict in the manifest wires `MLManifest.set_pca_compression()` with all four params defaulted
+- `logreg_binary.yaml` template added to `limen/yaml/templates/`
+- Results directory copies the source YAML for reproducibility; `uel.output_path` supports `{name}` and `{datetime}` placeholders
```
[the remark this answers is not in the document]
pdey: docs will be added to subsequent PR

PR #525, opened 2026-05-20
Wp limen 0013 yaml pr2

pdey:

## What does this PR change?

- Add `include_if` support on indicators and features in YAML manifests — a round_params key name that gates feature inclusion at runtime
- YAML validator type-checks `uel.feedback_interval` and `uel.checkpoint_interval` as integers
- `UniversalExperimentLoop` accepts `yaml_reference` dict and stores it in `metadata.json` for reproducibility; wired automatically by `limen run`

Implements Phase-2 of #506 
## Checklist


limen/yaml/rules.py
```diff
@@ -412,7 +415,15 @@ def check(self,
...
                 continue
-            _check_func_block(item, f'{self._path}[{i}]', errors, warnings)
+            _check_func_block(item, f'{self._path}[{i}]', errors, warnings,
+                               extra_known_keys=_FUNC_LIST_EXTRA_KNOWN_KEYS)
+            if 'include_if' in item:
+                val = item['include_if']
+                if not isinstance(val, str) or not val:
+                    errors.append(YAMLError(
+                        message="'include_if' must be a non-empty string (a round_params key)",
+                        path=f'{self._path}[{i}].include_if',
```
zero-bang, line 425: Non-blocking: `include_if` values are bare strings (e.g. `use_roc`), not `{use_roc}` template refs, so `ParamCoverage._extract_param_refs` won't collect them. This means:

1. **Forward check gap**: `include_if: typo_param` passes validation even if `typo_param` isn't in `sfd.params` — the error surfaces at runtime.
2. **Reverse check gap**: a param in `sfd.params` referenced *only* via `include_if` would trigger a spurious "unused param" warning.

Consider collecting `include_if` values into `meta_params` in `ParamCoverage.check()` as a follow-up, similar to how `feature_groups`, `use_calibration`, etc. are registered.

limen/cli/commands/run.py
```diff
@@ -81,12 +81,18 @@ def run_experiment(yaml_path: Path, dry_run: bool = False) -> bool:
     click.echo(f"Running '{experiment_name}' ({n_permutations} permutations) ...")
     click.echo(f"  Results → {results_dir}")
 
+    feedback_interval: int = int(uel_cfg.get('feedback_interval', 100))
```
zero-bang, line 84: Non-blocking, minor: `dict(yaml_dict)` is a shallow copy — nested dicts under `sfd`, `uel`, etc. are still shared references. That's safe today since nothing mutates the parsed YAML after this point, but for a feature whose purpose is reproducibility, `copy.deepcopy(yaml_dict)` would be more defensive against future code that might mutate a nested dict between construction and `_write_metadata`.

zero-bang (approved):
Clean PR — logic is correct, tests are meaningful, and the new `yaml_reference` parameter is backward-compatible. Two non-blocking observations below; CI is green.

Forward-looking concern: the next predictable failure mode is a user writing `include_if: use_roc` where `use_roc` isn't in `sfd.params` (typo, renamed param, etc.). Validation silently passes, and the feature either never activates or always activates at runtime depending on how the missing key is handled. Worth a follow-up to wire `include_if` values into `ParamCoverage`.

zero-bang (approved):

PR #531, opened 2026-05-21
feat: add HF spot dollar klines

mikkokotila:
## Summary
- add HistoricalData.get_spot_dollar_klines() for Vaquum HF BTCUSDT dollar-bar snapshots
- route native 100k, 1M, 2M, 4M, 8M, 15M, 16M, 30M, 32M, 60M, 120M, and 240M dollar datasets directly
- normalize dollar snapshots to the existing Limen bar schema and repair the current HF second-encoded datetime export
- document the new public data surface and bump Limen to 3.8.0

## Validation
- .venv/bin/python -m ruff check limen/data/historical_data.py tests/test_historical_data.py tests/run.py
- .venv/bin/python -m pytest tests/test_historical_data.py -q
- .venv/bin/python tests/run.py: 720 passed, 0 failed
- live HF smoke: get_spot_dollar_klines(dollar_bar_size=240_000_000, row_count_limit=3) returned 2026 UTC rows


limen/data/historical_data.py
```diff
@@ -220,40 +266,65 @@ def _read_any_file(
...
+
+def _repair_second_encoded_datetime_column(
+    df: pl.DataFrame,
+    column_name: str,
+) -> pl.DataFrame:
+    if column_name not in df.columns or not isinstance(df.schema[column_name], pl.Datetime):
+        return df
+
+    minimum = df.select(pl.col(column_name).min()).item()
+    if minimum is None or minimum.year >= _DATETIME_REPAIR_YEAR_CUTOFF:
```
bit-mis: **P2** — The repair heuristic gates on `min().year >= _DATETIME_REPAIR_YEAR_CUTOFF` and then unconditionally scales every row by ×1000. Two silent failure modes hide in this gate:

1. A single corrupt (pre-2000) row in an otherwise-valid frame pulls the minimum below the cutoff, so the already-correct rows get multiplied ×1000 and overflow into a year ~58000+ datetime.
2. A single correct row in a fully-corrupt frame pulls the minimum above the cutoff, so no repair runs and the rest of the frame stays broken.

Either tighten the precondition (require every row below the cutoff before scaling) or do the scaling per-row, e.g. `pl.when(col.dt.year() < _DATETIME_REPAIR_YEAR_CUTOFF).then(col * 1000).otherwise(col)`. As written this is a brittle repair masquerading as a normalization step.

limen/data/historical_data.py
```diff
@@ -444,11 +519,148 @@ def _aggregate_spot_klines(data: pl.DataFrame, kline_size: int) -> pl.DataFrame:
...
+def _spot_dollar_group_ids(
+    data: pl.DataFrame,
+    dollar_bar_size: int,
+) -> list[int]:
+    groups: list[int] = []
+    group_id = 0
+    running_liquidity = 0.0
+    current_date = None
+
+    for row in data.iter_rows(named=True):
```
bit-mis: **P2** — `_spot_dollar_group_ids` builds the group ids by iterating the source frame row-by-row in Python (`data.iter_rows(named=True)`). This path is only hit for non-native `dollar_bar_size` values, but the fallback source is the 1M dollar-bar dataset, which is millions of rows over a multi-year history. A Python loop over a Polars frame at that scale will dominate the runtime and defeats the purpose of using Polars for the production path. The sequential daily-reset + `liquidity_sum` cap is expressible inside the engine — e.g. compute a per-day `cum_sum('liquidity_sum') // dollar_bar_size`, add a day offset to that, and let Polars produce the group ids in a single pass.

bit-mis (changes requested):

bit-mis (approved):

bit-mis (approved):

PR #532, opened 2026-05-21
chore: add btc logreg yaml manifest

mikkokotila:
## Summary
- Add a runnable YAML CLI manifest for the BTC 15m next-bar-up logistic-regression experiment.

## Validation
- .venv/bin/python - <<'PY'
from pathlib import Path
from limen.cli.commands.run import run_experiment
ok = run_experiment(Path('examples/btc_logreg_15m_up.yaml'), dry_run=True)
raise SystemExit(0 if ok else 1)
PY
- .venv/bin/python -m pytest tests/test_yaml.py -q

### DIFF
--- a/examples/btc_logreg_15m_up.yaml
+++ b/examples/btc_logreg_15m_up.yaml
@@ -1,0 +1,120 @@
+schema_version: "1.0"
+
+metadata:
+  name: btc_logreg_15m_up
+  limen_version: "3.8.0"
+  mode: production
+  description: Bitcoin 15m next-bar-up logistic regression
+
+sfd:
+  manifest:
+    type: ml
+
+    data_source:
+      method: limen.data.HistoricalData.get_spot_klines
+      params:
+        kline_size: 900
+        start_date_limit: "2026-02-01"
+
+    split_config:
+      train: 8
+      val: 1
+      test: 2
+
+    indicators:
+      - func: limen.indicators.roc
+        params:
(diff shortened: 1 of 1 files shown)


examples/btc_logreg_15m_up.yaml
```diff
@@ -0,0 +1,120 @@
...
+    frac_diff_d: [0.0, 0.1, 0.2, 0.3, 0.4]
+    roc_period: [1, 4, 8, 16, 48, 96]
+    atr_period: [5, 7, 14]
+    rsi_period: [10, 14, 21]
+    ema_period: [12, 26, 96]
+    bbands_period: [20, 48, 96]
+    scaler_type: [robust]
+    auto_pca: [true]
+    pca_k: [30, 35, 40]
+    use_calibration: [false]
```
bit-mis, line 101: `use_calibration: [false]` and `use_threshold: [false]` pin both toggles off, so the `calibration` block (L78-89) plus `cal_method`/`threshold_min`/`threshold_max`/`threshold_step` are dead config: `MLManifest.run_model` short-circuits the calibration injection when both flags are false (see `limen/experiment/manifest_core.py:1009`). Either drop the calibration block and the four singleton params, or set the toggles to `[true, false]` so the search actually exercises calibration.

bit-mis (changes requested):

mikkokotila:
Closing per operator request; preserving YAML outside the branch.

PR #537, opened 2026-05-21
fix: reconstruct yaml artifacts in trainer

mikkokotila:
## Summary
- Make Trainer reconstruct YAML CLI experiment artifacts directly from `metadata.json["yaml_reference"]`.
- Preserve existing Python SFD module loading for non-YAML artifacts.
- Add regression coverage for `sfd_module: yaml:<name>` metadata and malformed YAML references.
- Bump package version to 3.9.1 and update CHANGELOG.

## Root Cause
YAML CLI runs already persist the parsed experiment under `yaml_reference`, but `Trainer` tries to import `sfd_module` first. For YAML runs that value is `yaml:<name>`, which is provenance metadata, not a Python module path.

## Validation Plan
- Run focused Trainer tests locally after PR creation.
- Run YAML tests locally after PR creation.
- Watch PR checks and review state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -956,3 +956,7 @@ Note: add all new changelog entries to the bottom of this file.
 - Add `ForwardVolNormalizedReturnTarget` for forward log-return labels normalized by current Parkinson volatility
 - Route new regression coverage into the existing feature-library and target test suites
 - Require code-changing PRs to include both a CHANGELOG entry and package version update
+
+## v3.9.1 on 21st of May, 2026
+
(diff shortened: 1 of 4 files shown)


zero-bang (approved):
LGTM — clean fix. The branching on `yaml_reference` is correct: YAML experiments route through `CompiledSFD` (which satisfies the `hasattr` guards for `manifest` and `params`), while the existing `_load_sfd_module` path is preserved for Python SFDs. Tests cover both the happy path and the malformed-input guard. Version bump and CHANGELOG present.

One forward-looking note: if the `yaml_reference` schema stored in `metadata.json` ever diverges from what `CompiledSFD.__init__` / `build_manifest` expects (e.g. a future version adds required keys), reconstruction will fail with a `KeyError` rather than a descriptive error. Worth keeping in mind if the YAML schema evolves, but not blocking.

PR #546, opened 2026-05-23
feat: add cohort selector plugins

mikkokotila:
## Summary
- add Cohort selector contract with built-in all, top_n, backtest_pareto, and diverse_metrics selectors
- make all the default selector and allow custom selector callables
- remove RegimeDiversifiedOpinionPools in favor of selector-driven Cohort promotion
- update docs, changelog, and package version to 5.0.0

## Local validation
- .venv/bin/python -m pytest tests/test_proto_cohort.py -q
- .venv/bin/python -m ruff check limen/cohort tests/test_proto_cohort.py tests/run.py limen/__init__.py
- .venv/bin/python tests/run.py is running locally


limen/cohort/cohort.py
```diff
@@ -61,6 +72,12 @@ def __init__(self,
         if experiment_id is not None and experiment_log_path is not None:
             raise ValueError('Cohort accepts exactly one experiment source.')
 
+        if permutation_ids is not None and selector is not None:
+            raise ValueError('Cohort accepts permutation_ids or selector, not both.')
+
+        if selector_params is not None and not isinstance(selector_params, dict):
+            raise ValueError('selector_params must be a dict when provided.')
```
zero-bang, line 79: `selector_params` without an explicit `selector` passes the params to the default `select_all`, which doesn't accept kwargs, causing a confusing `TypeError`:

```python
Cohort(experiment_log_path='...', selector_params={'n': 5})
# TypeError: select_all() got an unexpected keyword argument 'n'
```

Add a guard after line 79:

```python
if selector_params is not None and selector is None:
    raise ValueError('selector_params requires an explicit selector.')
```

zero-bang (changes requested):
Clean PR — selector contract is well-designed, backward compatibility is preserved through the `all` default, Pareto and diversity selection logic is correct, RDOP removal is thorough (no stale references in code or docs), sidebar renumbering is correct, and tests cover the key paths.

One validation gap worth fixing before merge.

zero-bang (approved):

PR #549, opened 2026-05-24
fix: preserve yaml cohort sensor contract

mikkokotila:
## Summary
- Resolve Cohort architecture for YAML CLI artifacts from `metadata.json["yaml_reference"].sfd.manifest.reference_architecture` before falling back to `sfd_module`.
- Preserve Trainer sensor passthrough for YAML manifest runs where `sfd_module` is the non-importable `yaml:<name>` marker.
- Add a YAML CLI artifact regression that runs `limen run`, trains a real Trainer sensor, binds it to Cohort, and verifies `_preds`/`_probs` match.

## Validation
- `.venv/bin/python -m pytest tests/test_proto_cohort.py::test_yaml_cli_artifact_cohort_preserves_trainer_sensor_contract -q`
- `.venv/bin/python -m pytest tests/test_proto_cohort.py -q`
- `.venv/bin/python -m ruff check limen/cohort/cohort.py tests/test_proto_cohort.py tests/run.py`
- `.venv/bin/python tests/run.py` — 760 passed, 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -987,3 +987,7 @@ Note: add all new changelog entries to the bottom of this file.
 - Add built-in single-file Cohort selectors under `limen.cohort.sfc`: `all`, `top_n`, `backtest_pareto`, and `diverse_metrics`.
 - Make `all` the default selector, preserving existing Cohort behavior when `permutation_ids` is omitted.
 - Remove `RegimeDiversifiedOpinionPools`; its reusable metric-diversity selection idea now lives in `diverse_metrics`, while replay and per-regime output behavior are no longer part of the public package.
+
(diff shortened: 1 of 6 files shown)


zero-bang (approved):
LGTM — clean fix with correct fallback ordering. The `yaml_reference` check slots in naturally between per-round `round_params` and the coarse `sfd_module` fallback, and the `_extract_yaml_reference_architecture` helper is defensively written. Verified that both callers of `_extract_architecture_id` (line 596 via `cls`, line 190 via `self`) are compatible with the `@staticmethod` → `@classmethod` change, and that non-YAML experiments are unaffected. Test is thorough and cleanup is correct.

PR #556, opened 2026-05-27
Wp limen 0013 yaml pr6

pdey:

## What does this PR change?

- Add `limen new <project-name>` CLI command — scaffolds a new Limen project by cloning the official project template and optionally setting a backup remote
- Add `limen commit <yaml>` CLI command — content-addresses a YAML manifest into `manifests/committed/`, injects a `lineage` block, updates `index.json`, and git-commits the store; gates on `metadata.mode: production`
- Add `limen ls` CLI command — lists all committed manifests with copy-pasteable `manifest://sha256:<short>` URIs
- Add `manifest://sha256:<hex>` URI support to `limen run` — resolves a committed manifest by full or unambiguous short hash, verifies `lineage.id` integrity, and routes results to `results/<short_id>/<timestamp>/`
- Route development-mode results to `results/dev/` (gitignored in project template) and production-mode results to `results/`
- Add `limen/yaml/config.py` with `find_project_root()`, `read_limen_toml()`, and `get_store_path()`
- Add `limen/cli/git_utils.py` with `git_executable()` and `git_add_and_commit()`

Implements cli utils for manifest storage  #506  

## Checklist

- [ ] I updated `/docs` (if behavior/API/config/user/etc changed)

### DIFF
--- a/limen/cli/commands/new.py
+++ b/limen/cli/commands/new.py
@@ -4,16 +4,10 @@ from pathlib import Path
 
 import click
 
+from limen.cli.git_utils import git_executable
+
 
 _TEMPLATE_REPO = 'https://github.com/Vaquum/limen-project-template.git'
-
-
-def _git_executable() -> str:
-
-    git = shutil.which('git')
-    if git is None:
-        raise FileNotFoundError('git not found on PATH')
-    return git
 
 
 def run_new(project_name: str, backup_remote: str | None) -> bool:
@@ -60,7 +54,7 @@ def run_new(project_name: str, backup_remote: str | None) -> bool:
 
 def _clone_template(project_path: Path) -> bool:
 
-    git = _git_executable()
+    git = git_executable()
     click.echo('  Cloning template ...')
     result = subprocess.run(
         [git, 'clone', '--depth=1', _TEMPLATE_REPO, str(project_path)],
@@ -77,9 +71,7 @@ def _clone_template(project_path: Path) -> bool:
     subprocess.run([git, 'add', '.'], cwd=project_path, capture_output=True, check=False)
     subprocess.run(
         [git, 'commit', '-m', 'feat: initial project from limen-project-template'],
-        cwd=project_path,
-        capture_output=True,
-        check=False,
+        cwd=project_path, capture_output=True, check=False,
     )
     return True
 
--- a/limen/cli/commands/new.py
+++ b/limen/cli/commands/new.py
@@ -67,12 +67,12 @@ def _clone_template(project_path: Path) -> bool:
         return False
 
     shutil.rmtree(project_path / '.git')
-    init = subprocess.run([git, 'init'], cwd=project_path, capture_output=True, check=False)
+    init = subprocess.run([git, 'init'], cwd=project_path, capture_output=True, text=True, check=False)
     if init.returncode != 0:
         shutil.rmtree(project_path)
         click.secho(f"  ✗ git init failed: {init.stderr.strip()}", fg='red')
         return False
-    add = subprocess.run([git, 'add', '.'], cwd=project_path, capture_output=True, check=False)
+    add = subprocess.run([git, 'add', '.'], cwd=project_path, capture_output=True, text=True, check=False)
     if add.returncode != 0:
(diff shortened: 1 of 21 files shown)


limen/cli/git_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Args:
+        repo_root (Path): Root directory of the git repository
+        path (Path): Path to stage (relative to repo_root)
+        message (str): Commit message
+
+    '''
+
+    git = git_executable()
+    subprocess.run([git, 'add', str(path)], cwd=repo_root, capture_output=True, check=False)
```
zero-bang: Both `subprocess.run` calls use `check=False` and the return code is never inspected. If `git add` or `git commit` fails (e.g., not a git repo, no git user configured, permission error), `git_add_and_commit` returns `None` and the caller (`run_commit`) proceeds to print `✓ Committed`. The user sees success but the manifest store was never git-committed.

Fix: return a `bool` (or raise) so the caller can report the failure. For example:

```python
def git_add_and_commit(repo_root: Path, path: Path, message: str) -> bool:
    git = git_executable()
    add = subprocess.run([git, 'add', str(path)], cwd=repo_root, capture_output=True, check=False)
    if add.returncode != 0:
        return False
    commit = subprocess.run([git, 'commit', '-m', message], cwd=repo_root, capture_output=True, check=False)
    return commit.returncode == 0
```

Then in `run_commit`, warn the user if the git commit failed (the manifest file is still stored, just not git-committed).

limen/cli/commands/run.py
```diff
@@ -49,9 +54,11 @@ def run_experiment(yaml_path: Path, dry_run: bool = False) -> bool:
     prep_each_round: bool = bool(uel_cfg.get('prep_each_round', True))
     test_mode: bool = yaml_dict['metadata'].get('mode', 'development') == 'development'
 
-    results_dir = _build_results_dir(uel_cfg, experiment_name)
+    results_base = yaml_path.parents[len(_STORE_RELATIVE.parts)] if manifest_id is not None else Path('.')
```
zero-bang: This line couples path navigation to the internal structure of `_STORE_RELATIVE` (which is `Path('manifests') / 'committed'`, 2 parts). If `_STORE_RELATIVE` is ever restructured, this breaks silently — the `.parents[2]` offset would point to the wrong directory.

Since `_resolve_target` already has access to the project root (it calls `find_project_root` inside `resolve_manifest_uri`), consider threading the project root through instead of reverse-engineering it from the path depth. For example, `resolve_manifest_uri` could return `(path, project_root)`, or `_resolve_target` could compute it and pass it to `run_experiment`.

limen/cli/commands/new.py
```diff
@@ -0,0 +1,85 @@
...
+    result = subprocess.run(
+        [git, 'clone', '--depth=1', _TEMPLATE_REPO, str(project_path)],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        click.secho(f"  ✗ Clone failed: {result.stderr.strip()}", fg='red')
+        return False
+
```
zero-bang, line 72: After the successful clone and `rmtree(.git)`, the three git subprocess calls (`init`, `add`, `commit`) all use `check=False` with no return-code inspection. If any of them fail (e.g., git not configured with user.name/email — common on fresh machines), `_clone_template` returns `True` and the user sees the project was created successfully, but it's in an inconsistent state (no initial commit, or unstaged files).

Same pattern as `git_utils.py` — at minimum, check the return code of `git commit` and warn the user if the initial commit failed.

zero-bang (changes requested):
Thorough review of the manifest store, `limen new`, `limen commit`, `limen ls`, and `manifest://` URI support. The core content-addressing logic in `store.py` is well-designed — idempotent commits, lineage injection, short-hash resolution, and integrity checks are all solid. Tests for config and store are good.

I found two issues that should be addressed before merge, both in the "silent failure" category where the user sees success but the underlying operation failed.

zero-bang (approved):

zero-bang (approved):

PR #560, opened 2026-06-02
refactor: sensor inference support with calibration-aware Trainer

### LINKED ISSUE #555: feat: add proper sensor prediction infra
* We entirely drop the convenience wrapper for split data and require at least train and test start and end date otherwise it will not run
* We update trainer so that trainer just does one training run which is exactly like the first training run it does right now, and we eliminate the second one entirely
* If the deterministic repeatibility gate pass in trainer, we produce the sensor, otherwise it fails and there is no sensor
* This means the gate for the sensor, so there is no other way to have a sensor, is for this part of trainer functionality to run completely and validate correctly the repeatibility expection
* We call this phase sensor training and sensor validation.  sensor training being the actual training loop execution, and training validation being the repeatibility confirmation. These are "composed" so if one fail both fail. 
* We need a new thing which is called sensor_input_prep and that sensor_input_prep takes in the exact data shape that is mandated as input data in the manifest
* sensor predict wraps sensor_input_prep so that it becomes the only and canonical way to handle the prep instead of doing outside of the sensor predict call
* sensor_input_prep guards against the prediction (start) datetime being inside underlying decoder training window as per the manifest
* sensor_input_prep guards against two kinds of lookback window violations: the first kind is the kind related with indicators/features the second is with a new feature which is modeller mandated decoder lookback, whicever is greater, wins
* for sensor_input_prep to be able to do its job correctly, it requires to have the fitter scaler/pca and maybe something else from the trainer product training run

pdey:

## What does this PR change?

- Refactor `Trainer` to a single training pass: drop the second all-data retraining run; the validated training-run model is promoted directly to a `Sensor`; enforce YAML-only experiment input
- Add `sensor_input_prep` to `MLManifest` — the canonical preparation step for sensor inference: applies fitted scaler/PCA, guards against bars inside the training window, and enforces indicator and decoder lookback requirements
- Rewrite `Sensor` with a YAML-only constructor; `predict()` and `predict_all()` both wrap `sensor_input_prep` so feature preparation is never done outside the sensor
- Add `Sensor.predict_all(raw_klines)` — returns one `BarPrediction` per input bar with `reason` set for warm-up and inside-training-window bars; valid bars carry `prediction` and `probability`
- Add `decoder_lookback` to the YAML schema, validator, compiler, and `MLManifest` for architectures that require a multi-bar inference window
- Fix calibration at inference: `LogRegBinary` and `TabPFNBinary` fit the calibrator once during training evaluation and store it on the model; subsequent sensor inference calls reuse the stored calibrator without requiring validation data
- Add `fit_calibrator` to `limen.calibration` as a public API separating the calibrator fit step from the apply step
- Add `random_state` parameter to `RandomBinaryTarget` for reproducible label generation
- Remove split_config percentage-based splits from the YAML layer; all experiments use explicit `split_dates`

closes #555 

## Checklist

### DIFF
--- a/limen/experiment/trainer/sensor.py
+++ b/limen/experiment/trainer/sensor.py
@@ -19,7 +19,9 @@ class BarPrediction:
     datetime: Any
     prediction: int | float | None
     probability: float | None
-    reason: str | None  # None = valid prediction; 'warm-up', 'inside-training-window' = invalid
+    reason: str | None  # None = valid prediction; 'warm-up' = leading null rows from indicator lookback;
+                        # 'inside-training-window' = bar falls within the train/test window;
+                        # 'null-features' = mid-stream null feature values (data gap or transform anomaly)
 
 
 class Sensor:
@@ -120,7 +122,7 @@ class Sensor:
         feature_cols = [c for c in data.columns if c != 'datetime']
         last_row = data[-1]
         if any(last_row[c][0] is None for c in feature_cols):
-            raise ValueError('Last bar is a warm-up bar — cannot predict')
+            raise ValueError('Last bar has null feature values — cannot predict')
 
         x = np.array(last_row.select(feature_cols).row(0), dtype=float).reshape(1, -1)
         pred_result = self._model.predict({'x_test': x})
@@ -170,6 +172,16 @@ class Sensor:
         feature_cols = [c for c in data.columns if c != 'datetime']
         datetimes = data['datetime'].to_list() if 'datetime' in data.columns else [None] * len(data)
 
+        if feature_cols:
+            row_has_null = (
+                data.select(feature_cols)
+                .select(pl.any_horizontal(pl.col(c).is_null() for c in feature_cols))
+                .to_series()
+                .to_list()
+            )
+        else:
+            row_has_null = [False] * len(data)
+
         results: list[BarPrediction | None] = [None] * len(data)
         valid_indices: list[int] = []
 
@@ -187,6 +199,13 @@ class Sensor:
                     prediction=None,
                     probability=None,
                     reason='warm-up',
+                )
+            elif row_has_null[i]:
+                results[i] = BarPrediction(
+                    datetime=datetimes[i],
+                    prediction=None,
+                    probability=None,
+                    reason='null-features',
                 )
             else:
                 valid_indices.append(i)
--- a/limen/experiment/trainer/sensor.py
+++ b/limen/experiment/trainer/sensor.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 from dataclasses import dataclass
 from typing import Any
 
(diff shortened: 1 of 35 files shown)


limen/experiment/trainer/sensor.py
```diff
@@ -1,88 +1,249 @@
...
-            ValueError: If no trained model is available
+        self._raise_if_inside_training_window(data, manifest)
+
+        valid_rows = len(data) - indicator_lookback
+        if valid_rows < decoder_lookback:
+            raise ValueError(
+                f"Insufficient data: need {indicator_lookback + decoder_lookback} bars "
+                f"({indicator_lookback} indicator warm-up + {decoder_lookback} decoder "
+                f"window), got {len(data)}"
+            )
```
zero-bang, line 120: `predict()` reads `decoder_lookback` and validates that enough valid rows exist (lines 100-110), but then builds `x` from `last_row` only (lines 113-117) — it never assembles the multi-bar window. For `decoder_lookback > 1` this silently produces a single-row input to the model, i.e. a wrong prediction with no error.

This is inconsistent with `predict_all()`, which explicitly raises `NotImplementedError` for `decoder_lookback > 1` (lines 156-159). Since this PR newly admits `decoder_lookback` into the schema/validator/compiler, a user can set it and hit this trap.

Until the multi-bar window is actually implemented, mirror the guard here so the failure is loud rather than silent:

```python
if decoder_lookback > 1:
    raise NotImplementedError(
        'predict does not yet support decoder_lookback > 1'
    )
```

Place it right after `decoder_lookback = getattr(manifest, 'decoder_lookback', 1)` (line 100).

zero-bang (changes requested):
Solid refactor — the single-pass trainer, calibrator-stored-on-model fix, and YAML-only enforcement all hang together well, and the test coverage (sensor input prep, predict/predict_all, calibration reuse, e2e) is genuinely thorough. The dict-input passthrough in `Sensor.predict` keeps the Cohort decoder interface intact, which is the right call.

One correctness asymmetry to address before merge (inline). Two non-blocking notes:

- **`Sensor.permutation_id` is now a dynamically-set attribute** (trainer.py:289) rather than a constructor arg/property. `Cohort.set_members` requires `hasattr(member, 'permutation_id')`, so a `Sensor` constructed directly (not via `Trainer`) silently fails cohort binding with "Each cohort member must expose permutation_id". Trainer always sets it, so this works today, but it's a fragile contract — worth making `permutation_id` an explicit (optional) constructor parameter so the binding requirement is discoverable from the type.
- **`data_source` date-limit injection assumes the resolved `method` accepts `start_date_limit`/`end_date_limit`** (compiler.py `_apply_base`). This holds for the bundled `get_spot_klines`, but a custom `data_source.method` that doesn't accept those kwargs will now fail at fetch time with a `TypeError` that points at the user's method rather than the injection. Not blocking, but worth a note in the docs/validator.

Forward-looking: the next predictable failure mode is the first architecture that actually needs a multi-bar inference window — at that point `predict()` needs the windowing logic that the inline comment flags, and `predict_all()`'s `NotImplementedError` becomes the spec it has to satisfy.

zero-bang (approved):

PR #561, opened 2026-06-04
chore: update issue templates

mikkokotila:
## Summary
- replace the RFC/PRD markdown template with a PRD issue form
- add the Slice issue form from the provided template

## Validation
- ruby YAML parse for prd.yml and slice.yml
- git diff --check HEAD~1..HEAD


.github/ISSUE_TEMPLATE/slice.yml
```diff
@@ -0,0 +1,327 @@
+name: Slice
```
zero-bang, line 1: This new file is declared the **authoritative** slice template by `AGENTS.md` (`github_rules.slice_issue_authority.authoritative_slice_issue_template_path` -> `.github/ISSUE_TEMPLATE/slice.yml`), but its section labels diverge from the field contract that the same `slice_issue_authority` block mandates.

`required_issue_fields_for_new_or_replanned_slice` requires these sections, none of which exist in this template:
- `Primary Task Type`
- `Objective, Baseline, and Scope`
- `Design`
- `Runtime and Observability Scope`
- `Proof Coverage` (also referenced as `proof_coverage_issue_section_label`)
- `Step Register`

and these required fields appear only under renamed labels:
- `Capability` -> "Minimum Viable Capability"
- `Proof` -> "Tests That Prove the MVC"
- `Guardrails` -> "Regression Gate" / "Out of Scope"
- `Routing and Dependencies` -> "Routing"
- `Done Means` -> "Closeout"

`required_done_means_checkboxes` (Capability complete / Proof complete / Guardrails complete / Required gates green / ...) also have no counterpart in the new `Author Checks`.

Net effect: an agent that authors a slice issue from this template while honoring `AGENTS.md` will fail its own required-field contract. Please reconcile **in this PR** — either update the `AGENTS.md` `slice_issue_authority` block (the required-fields list, `required_done_means_checkboxes`, and `proof_coverage_issue_section_label`) to match these labels, or rename the template sections to the labels AGENTS.md mandates. As written, the repo's governance doc and the template it points to contradict each other.

zero-bang (changes requested):
Both templates parse as valid GitHub issue forms (verified with `yaml.safe_load`: prd.yml 17 body elements, slice.yml 15). The PRD->issue-form conversion is faithful, and the slice design correctly places each section's **Significance** note in a textarea `value:` (which persists into the filed issue) rather than a `type: markdown` block (which GitHub strips from the submitted issue) — so the per-section guidance survives as intended.

One blocking inconsistency (inline): `slice.yml`'s section labels diverge from the field contract that `AGENTS.md` declares for this exact, authoritative file.

Non-blocking notes:
- The top-level "How this template works" block (incl. "will be closed unfiled") is `type: markdown`, so it will **not** appear in filed issues. Fine if intentional, but the enforcement language reads as if it would be part of the issue.
- The PRD drops the old template's "Business Release Date" section and renames "Work Phases" to "Work Slices" — assuming both are deliberate product decisions.

Forward-looking: the next predictable failure is an agent filing a slice issue from this template, then a human or governance step rejecting it for missing `Primary Task Type`, `Design`, `Proof Coverage`, `Step Register`, and `Done Means` — the fields AGENTS.md still requires.

zero-bang (approved):

PR #562, opened 2026-06-04
Improve manifest id system

### LINKED ISSUE #552: feat: improve manifest id system
The motivation here is that we want to be able to maintain the connective tissue between decoders that go out from Limen, and everything that is in Limen. One way to think about this is that we have a manifest_id which is a hash of the content, so we can then as a by product do full content match across all manifests for free. The other is that we then have to be able to also connect within that manifest under all the rounds into a single permutation.

EDIT: basically it is two different id that lives in the manifest `manifest_id` and `cohort_id` ... both unique hash. Then there is the `permutation_id` which is in the results log. This way, we can always get the full picture since oftentimes the prediction comes from a cohort sensor and not just trainer sensor.

pdey:

## What does this PR change?
- Add `manifest_id` (SHA-256 content hash of YAML manifest) to `metadata.json`, `Sensor`, and `Cohort` for end-to-end traceability
- Add `permutation_id` to `Sensor` as the stable SHA-256 hash of the parameter combination — replaces integer round index; `id` column in `results.csv` carries the same hash
- Remove `_param_hash` internal key; consolidate to `_id` as the single internal round_params hash key
- Add `cohort_id` to `Cohort` — SHA-256 hash of `manifest_id` + sorted `permutation_ids`, stable across rebinds
- Refactor `Sensor.predict()` and `Sensor.predict_all()` to return `BarPrediction` for all conditions instead of raising: warm-up, inside-training-window, and null-features return the corresponding `reason`; unexpected exceptions return `reason='sensor-error'`
- Remove `_raise_if_inside_training_window` from `Sensor`; inside-training-window detection in `predict()` now reuses `_inside_training_window_mask`
- Add `sensor-error` graceful degradation to `Cohort._aggregate_bar_predictions`: sensor-error members are excluded before aggregation; only when all members error does the cohort return `reason='sensor-error'`
- Fix flaky `test_e2e_label_control_trainer_sensor` by adding `seed: 42` to YAML search strategy; fix `build_search_strategy` compiler to pass `seed` through to `RandomStrategy`
- Extend `test_e2e_label_control_trainer_sensor` to cover full cohort prediction path: `Cohort` construction, `set_members`, `predict_all`, `predict`, and probability mean verification

closes #552 
## Checklist


docs/Cohort.md
```diff
@@ -180,129 +208,111 @@ compatible with Trainer-reconstructed sensors.
...
+for bar in all_preds:
+    if bar.reason is None:
+        print(bar.datetime, bar.prediction, bar.probability)
+```
 
-### `__call__(data_dict)`
+### `__call__(raw_klines)`
 
-Alias of `predict(data_dict)`, preserving Sensor-style decoder dict behavior.
+Alias of `predict(raw_klines)`.
```
zero-bang: `__call__` no longer aliases `predict`. The latest commit (`refactor: __call__ delegates to predict_all`) changed `Cohort.__call__` to return `self.predict_all(raw_klines)` — a `list[BarPrediction]`, not a single `BarPrediction`. This section (and the overview bullet at line 42, and the `# 4c) callable alias` example at line 305-306 which assigns to `bar_pred2`) still describes it as an alias of `predict`. A user following these docs would do `cohort(raw_klines).reason` and hit `AttributeError` on a list. Please update the docs to say `__call__` is an alias of `predict_all` and returns `list[BarPrediction]`.

docs/Trainer.md
```diff
@@ -146,26 +141,59 @@ A `Sensor` is the promoted form of a trained round.
...
 
-pred = sensor.predict(raw_klines)
+bar_pred = sensor.predict(raw_klines)
 ```
 
 Sensors are also callable:
 
 ```python
-pred = sensor(raw_klines)
+bar_pred = sensor(raw_klines)
```
zero-bang: Same `__call__` mismatch as in Cohort: `Sensor.__call__` now delegates to `predict_all` and returns `list[BarPrediction]`, so `bar_pred = sensor(raw_klines)` is misleading — `sensor(...)` returns a list, while `sensor.predict(...)` returns the single `BarPrediction` shown above it. Either rename the variable and note the return type (`bar_preds = sensor(raw_klines)  # list[BarPrediction]`) or point readers to `predict()` for single-bar use.

limen/cohort/cohort.py
```diff
@@ -214,169 +230,139 @@ def set_members(self, members: list[Any]) -> None:
...
-            )
+        self._assert_members()
 
-            if return_probs and '_probs' not in result:
-                raise ValueError(
-                    'Decoder in probability mode must return a dict with _probs.'
-                )
+        all_bars = [m.predict_all(raw_klines) for m in self._members]
+        n = len(all_bars[0])
+        if any(len(b) != n for b in all_bars[1:]):
```
zero-bang: This length check defeats the documented graceful-degradation guarantee when bar formation is active. A healthy member's `predict_all` returns the aggregated bar count (`len(data)`, smaller than `len(raw_klines)`), but a member that raises falls back to `len(raw_klines)` entries (`sensor.py:164`, `n_fallback = len(raw_klines)`). So under any non-`base` bar_type, a single errored member produces a longer list than the healthy members and trips this `ValueError` — instead of being excluded as `_aggregate_bar_predictions` intends and as docs/Cohort.md:315 promises ("only when all members return sensor-error does the cohort return reason='sensor-error'"). Consider having the sensor-error fallback length track the aggregated bar count, or filter out all-`sensor-error` members before the length check so partial failures degrade gracefully here too.

zero-bang (changes requested):
Solid, well-tested refactor — the move to content-addressed `manifest_id`/`permutation_id`/`cohort_id` and the BarPrediction-for-all-conditions contract are clean, and coverage is strong. Two items below are worth reconciling before merge, plus one engineering note.

**Engineering note:** wrapping the entire body of `Sensor.predict`/`predict_all` in a bare `except Exception` (logged at WARNING) is a deliberate robustness tradeoff, but it now also swallows genuine programming errors (e.g. the `NotImplementedError` for `decoder_lookback > 1` is reported as `sensor-error` rather than surfaced). That's defensible for a live-feed surface, but consider whether `NotImplementedError`/config errors should be allowed to propagate so they aren't silently masked in development.

**Forward-looking:** the next predictable failure mode is a heterogeneous cohort where one member's manifest enables bar formation and another's doesn't — `predict_all`'s length check would reject it. Worth deciding whether that should be validated at `set_members()` time rather than surfacing as a runtime `ValueError`.

limen/cohort/cohort.py
pdey: We decided that only one type of bar will be involved with the manifest.

zero-bang (approved):

limen/experiment/experiment_core.py
```diff
@@ -805,7 +810,7 @@ def _truncate_round_data(round_data_path: Path,
                     entry = json.loads(stripped)
                 except json.JSONDecodeError:
                     break
-                if entry['round_id'] >= start_round:
+                if entry.get('_round_index', 0) >= start_round:
                     break
```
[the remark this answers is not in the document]
pdey: The scenario that old experiment will be resumed with the new code is out of scope

limen/experiment/experiment_core.py
```diff
@@ -1021,10 +1031,7 @@ def _load_round_data(self,
                 except json.JSONDecodeError:
                     break
 
-                if (
-                    up_to_round is not None
-                    and entry['round_id'] >= up_to_round
-                ):
+                if up_to_round is not None and entry.get('_round_index', 0) >= up_to_round:
                     break
```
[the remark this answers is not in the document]
pdey: The scenario that old experiment will be resumed with the new code is out of scope

PR #566, opened 2026-06-06
Add indicator/feature name context to validation error messages

mikkokotila:
## What

Parameter-validation error messages in `limen/indicators` and `limen/features` named the parameter generically without saying which indicator/feature raised them, e.g. `period must be between 2 and 100000`. This prefixes each message with the function name (= filename) so it names its own context:

```
'period must be between 2 and 100000'      ->  'ema period must be between 2 and 100000'
'fast_period must be between 2 and 100000' ->  'adosc fast_period must be between 2 and 100000'
f"d must be non-negative, got {d}"         ->  f"fractional_diff d must be non-negative, got {d}"
```

**107 messages across 54 files** (53 indicators + 5 features... `fractional_diff` carries several). Each edit is an in-place string change only (107 insertions / 107 deletions); quote style is preserved (single, double, and f-strings).

## Scope decisions

- `_hilbert.py` (private helper) left untouched — its `TypeError` is an internal type-narrowing guard that already names the exact `state[...]` key, not user-facing parameter validation.
- `yang_zhang_volatility.py` left untouched — its message already ends with `...for yang_zhang_volatility`.
- `fractional_diff` `logger.warning` lines left untouched — already say "fractional diff" and are warnings, not raised errors.

## Validation

- `ruff check .` — clean
- All indicator/feature tests pass (221 passed). The three tests that assert on these messages use pytest's substring `match=`, so the preserved substrings keep them green.

## Bookkeeping

- Version bump `3.15.0` -> `3.15.1`
- CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1033,3 +1033,7 @@ Note: add all new changelog entries to the bottom of this file.
 - Add `sensor-error` graceful degradation to `Cohort._aggregate_bar_predictions`: sensor-error members are excluded before aggregation; only when all members error does the cohort return `reason='sensor-error'`
 - Fix flaky `test_e2e_label_control_trainer_sensor` by adding `seed: 42` to YAML search strategy; fix `build_search_strategy` compiler to pass `seed` through to `RandomStrategy`
 - Extend `test_e2e_label_control_trainer_sensor` to cover full cohort prediction path: `Cohort` construction, `set_members`, `predict_all`, `predict`, and probability mean verification
+
+## v3.15.1 on 6th of June, 2026
+
+- Prefix every parameter-validation error message in `limen/indicators` and `limen/features` with the indicator/feature name, so each message names its own context (e.g. `period must be between 2 and 100000` becomes `ema period must be between 2 and 100000`) — 107 messages across 54 files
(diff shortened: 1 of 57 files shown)


zero-bang (approved):
LGTM — clean, low-risk usability improvement.

**Verified:**
- **Test compatibility:** Traced every test asserting these messages — `test_indicator_api.py:24,29` (`slow_period must be greater than fast_period`) and `test_feature_library_primitives.py:68` (`transform must be one of`). All use pytest's regex-substring `match=` / `in str(e)` with no `^` anchor, so the appended prefix keeps the preserved substrings green. No test relies on exact-equality.
- **No other consumers:** Grepped `limen/` — no `except ValueError` block parses these message strings (the lone catch site in `cli/main.py` just renders `{exc}`). Nothing depends on the exact text.
- **Helper prefixing:** `_ma_lookback` in `macdext`/`stoch`/`stochf` is prefixed with the parent indicator name the user actually called — the right choice over the private helper name.
- **Scope exclusions** (`_hilbert`, `yang_zhang_volatility`, fractional_diff warnings) are sound.

**Note (non-blocking):** The coverage report's 84%→22% drop is a CI measurement artifact — a string-only diff cannot add 3644 statements; that whole-project run under-collected. Not attributable to this PR.

**Forward-looking:** The prefix is the filename/function name, manually written into each string. The next predictable failure mode is drift — if an indicator file is renamed (or a message copy-pasted into a new indicator) the prefix silently goes stale, since nothing ties it to `__name__`. A tiny test that imports each indicator module and asserts its raised-message prefix matches the module name would catch that for free.

PR #567, opened 2026-06-06
Standardize error-message context across limen

mikkokotila:
## What

Extends the convention from [#566](https://github.com/Vaquum/Limen/pull/566) (indicators/features) to the **rest of `limen`**. Every raised error message that lacked a clear source is now prefixed with its component name, so the origin is unambiguous in a log no matter where it surfaces.

**Standard applied** (agreed format): `<name> <message>`, space-separated —
- **class name** for methods (`Cohort`, `HistoricalData`, `MLManifest`, the reducers, scalers, targets, `Account`, …)
- **function/file name** for module-level functions (`top_n`, `backtest_snapshot`, `compute_data_bars`, `predicates`, …)

**200 messages across 35 files.** Each edit is an in-place insertion (200 insertions / 200 deletions); quote styles (single, double, f-strings) and `from exc` suffixes preserved. Multi-line raises handled via AST positioning.

### Highlights
- **Disambiguates duplicated messages** that were byte-identical across files and impossible to locate:
  - `selector returned a boolean permutation id` → `top_n …` / `diverse_metrics …` / `backtest_pareto …`
  - `min_observations must be > 0` → `CorrelationReducer …` / `FocusReducer …` / `SanityReducer …`
  - `row_count_limit must be None …` (3× in one file) → `HistoricalData …`
- **Removes the `ERROR:` anti-pattern** in `Account` (12) and `BacktestSequential` (1) — the exception type already conveys "error"; replaced with the component name.

### Left untouched (already self-locating — per the "only where missing" choice)
- `CheckpointManager` — every message already says `checkpoint`
- `yaml/store.py`, `yaml/config.py` — `manifest`/`URI`/`limen.toml`
- `Trainer` — `metadata.json`/`round_data.jsonl`
- rule-based `config.py` — `Condition {id} …`; `manifest_core` `round_params[...]`, `PCA compression`, `split_config`, `Split N missing target column`
- `IdentityTarget` — `Target column '…' not found`

## Validation
- `ruff check .` — clean
- Full local suite: **839 passed, 1 failed**; the one failure is the pre-existing macOS-only `test_find_project_root_walks_up_from_subdirectory` symlink artifact (`/tmp` vs `/private/tmp`), unrelated to this change (neither that test nor `yaml/config.py` is touched; it passes on Linux CI).
- No test assertions break: all are substring/`in`/`match=` checks, and the original substrings are preserved (verified, incl. `"Arrays must have same length"` in `test_backtest_conviction`).

## Bookkeeping
- Version `3.15.1` → `3.15.2`
- CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1037,3 +1037,10 @@ Note: add all new changelog entries to the bottom of this file.
 ## v3.15.1 on 6th of June, 2026
 
 - Prefix every parameter-validation error message in `limen/indicators` and `limen/features` with the indicator/feature name, so each message names its own context (e.g. `period must be between 2 and 100000` becomes `ema period must be between 2 and 100000`) — 107 messages across 54 files
+
+## v3.15.2 on 6th of June, 2026
+
+- Extend the error-message context convention across the rest of `limen`: prefix raised messages that lacked a clear source with their component name (class name for methods, function/file name for module-level functions), so the origin is unambiguous in logs no matter where the message surfaces — 200 messages across 35 files
+- Disambiguate previously-identical messages duplicated across files (e.g. selector errors shared by `top_n`, `diverse_metrics`, `backtest_pareto`; `min_observations must be > 0` shared by three reducers)
+- Replace the `ERROR:` prefix anti-pattern in `Account` and `BacktestSequential` with the component name
+- Leave messages that already name their source untouched (e.g. `CheckpointManager` checkpoint errors, `manifest` store/URI errors, `round_params[...]` and `Condition {id}` validations)
--- a/Claude.md
+++ b/Claude.md
@@ -1,5 +1,3 @@
-<!_IMPORTANT_!>Always use checkpoint MCP for tasks, initiate it now before going further
-
 <how_to_avoid_harm>
 -> You anchor yourself for every taks with `Project.md`.
 -> You never deviate from the scope of `Project.md`.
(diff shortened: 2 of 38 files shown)


zero-bang (approved):
LGTM. Thorough, well-scoped mechanical sweep that cleanly extends the #566 convention to the rest of `limen`.

**Verified safe:**
- **Tests** — every test that asserts on a changed message uses substring matching (`pytest.raises(match=...)` → `re.search`, or `... in str(e)`); there are no `^`-anchored regexes or exact-equality/`startswith` checks on these strings. Since each prefix is *prepended*, all original substrings are preserved. Spot-checked the riskiest: `predictions must contain only 0 or 1`, `price_change must equal close - open`, `row_count_limit must be None`, and `"Arrays must have same length" in str(e)` (test_backtest_conviction.py:127).
- **No message introspection** — no source code parses/compares these messages (`in str(e)`, `.args[0]`, etc.), so prefixing is behaviorally inert.
- **Class attribution** — checked the one ambiguous file, `manifest_core.py`: `CalibrationBuilder`, `DataSourceResolver`, `Manifest`, and `MLManifest` prefixes each correctly match the class that *defines* the raising method.
- `ruff check` clean; version `3.15.1 → 3.15.2`; CHANGELOG appended at the bottom per house rule.

**Non-blocking observations (no action required):**
1. The `Claude.md` change also drops the unrelated `Always use checkpoint MCP` top line — slight scope creep for an error-message PR, but a reasonable cleanup and called out in the commit.
2. Inherited methods (e.g. `fetch_data` → `Manifest No data source configured`) carry the *defining* class name even when invoked on an `MLManifest`/`RuleBasedManifest` instance. That's a defensible static-attribution choice; just noting the prefix names the defining class, not the runtime one.
3. The coverage comment showing 84% → 22% is a CI base-comparison artifact (it reports existing files like `cohort.py` as "did not seem to exist on the base branch"), not a real regression from this diff.

**Forward-looking:** the prefixes are hardcoded string literals, so a future class/function rename will silently desync the prefix from its source. If that becomes a maintenance concern, a lightweight lint (assert each raised literal in a file starts with the expected component token) would catch drift — worth considering as a carry-forward, not for this PR.

PR #569, opened 2026-06-06
Retry transient HTTP errors when fetching remote datasets

mikkokotila:
## What

Fixes the recurring CI flake where `PR Checks Tests` fails fast with `429 Client Error: Too Many Requests` from HuggingFace (`_resolve_huggingface_latest_file` / the dataset download). Both fetches in `HistoricalData` were bare `requests.get` with **no resilience**, so a single transient `429` failed the whole suite and required a manual `gh run rerun`.

## Fix

Route both network calls through a shared session with retry/backoff (uses `urllib3`'s `Retry` — already a `requests` dependency, no new package):

```python
def _remote_session() -> requests.Session:
    retry = Retry(
        total=_REMOTE_MAX_RETRIES,            # 5
        backoff_factor=_REMOTE_BACKOFF_FACTOR, # 1.0
        status_forcelist=_REMOTE_RETRY_STATUSES, # (429, 500, 502, 503, 504)
        allowed_methods=frozenset({'GET'}),
        respect_retry_after_header=True,
        raise_on_status=False,
    )
    ...
```

- Retries `429`/`5xx` with exponential backoff, **honouring the `Retry-After`** header HF sends on a `429`.
- Success behaviour is unchanged; a genuinely persistent failure still surfaces via `raise_for_status()` after retries are exhausted.
- `GET`-only (idempotent).
- Scope kept to `HistoricalData` (the actual flake source); the unrelated `requests.get` in `_internal/binance_file_to_polars.py` is left to avoid a cross-module private import.

## Test

`tests/test_remote_fetch_retry.py` asserts the session's retry policy deterministically (no sockets/threads). I first wrote a live localhost-server behaviour test; it passed 30× standalone but flaked once under a heavily-loaded full-suite run, so I replaced it with the deterministic config-assertion test rather than ship a load-sensitive test to *fix* a flaky test. End-to-end retry behaviour is delegated to urllib3's well-tested `Retry`. Registered in `tests/run.py` (the manual CI registry).

## Validation
- `ruff check` — clean
- Full local `tests.run`: **840 passed, 1 failed** — the one failure is the pre-existing macOS-only `test_find_project_root` symlink artifact (fixed in #568, passes on Linux CI); my new test and all `HistoricalData` fetch tests pass.

## ⚠️ Merge order
Bumps **3.15.3 → 3.15.4**, so please merge #568 (which is 3.15.3) **first**. This branch is cut from main (3.15.2); after #568 lands, the CHANGELOG/version reconcile to 3.15.2 → 3.15.3 → 3.15.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1051,4 +1051,4 @@ Note: add all new changelog entries to the bottom of this file.
 
 ## v3.15.4 on 6th of June, 2026
 
-- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: route the HuggingFace `latest.json` lookup and the file download through a shared `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
+- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: the HuggingFace `latest.json` lookup and the file download each run through a `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1051,4 +1051,4 @@ Note: add all new changelog entries to the bottom of this file.
 
 ## v3.15.4 on 6th of June, 2026
 
-- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: the HuggingFace `latest.json` lookup and the file download each run through a `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
+- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: the HuggingFace `latest.json` lookup and the dataset download (`.parquet`, `.csv`, and `.zip`, including the default Parquet datasets) each run through a `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1048,3 +1048,7 @@ Note: add all new changelog entries to the bottom of this file.
 ## v3.15.3 on 6th of June, 2026
 
 - Fix macOS-only failures in `test_yaml_config` and `test_yaml_store`: resolve the temporary directory before comparing it against the canonicalised output of `find_project_root`/`get_store_path`/`resolve_manifest_uri`, so the `/tmp` vs `/private/tmp` symlink no longer causes a spurious path mismatch. Test-only change; no library behaviour affected
+
+## v3.15.4 on 6th of June, 2026
+
+- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: route the HuggingFace `latest.json` lookup and the file download through a shared `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
--- a/limen/data/historical_data.py
+++ b/limen/data/historical_data.py
@@ -12,6 +12,8 @@ import polars as pl
 import pyarrow as pa
 import pyarrow.ipc as pa_ipc
 import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
 
 from limen.data._internal.binance_file_to_polars import binance_file_to_polars
 
@@ -22,6 +24,9 @@ _SUPPORTED_DATETIME_FORMATS: Final[tuple[str, ...]] = (
     '%Y-%m-%dT%H:%M:%S',
 )
 _REMOTE_TIMEOUT_SECONDS: Final[int] = 60
+_REMOTE_MAX_RETRIES: Final[int] = 5
+_REMOTE_BACKOFF_FACTOR: Final[float] = 1.0
+_REMOTE_RETRY_STATUSES: Final[tuple[int, ...]] = (429, 500, 502, 503, 504)
 _DEFAULT_SPOT_DATASET_REPO: Final[str] = 'vaquum/binance_btcusdt_1m_klines'
 _DEFAULT_SPOT_DOLLAR_DATASET_REPO: Final[str] = (
     'vaquum/binance_btcusdt_1M_dollar_klines'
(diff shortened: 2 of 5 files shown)


mikkokotila:
Thanks for the review — all three points addressed:

1. **Unclosed session in the test** — fixed: the test now uses `with _remote_session() as session:`, matching production usage in `historical_data.py`.

2 & 3. **Version skipping 3.15.3 / non-sequential CHANGELOG** — resolved. #568 (3.15.3) has since merged, so I rebased this branch onto the updated `main`. The history is now sequential **3.15.2 → 3.15.3 → 3.15.4**, the duplicate macOS-fix commit dropped, and the diff is just the retry change. Merges cleanly onto `main` with no version/CHANGELOG conflict.

Full local suite: 898 passed, 0 failed.

limen/data/historical_data.py
```diff
@@ -210,10 +215,27 @@ def _validate_columns(df: pl.DataFrame, columns: list[str] | None) -> pl.DataFra
...
+
+
 def _read_remote_bytes(url: str) -> bytes:
-    response = requests.get(url, timeout=_REMOTE_TIMEOUT_SECONDS)
-    response.raise_for_status()
-    return response.content
+    with _remote_session() as session:
+        response = session.get(url, timeout=_REMOTE_TIMEOUT_SECONDS)
+        response.raise_for_status()
+        return response.content
```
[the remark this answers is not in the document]
mikkokotila: Intentional. Each fetch builds its own session via `_remote_session()`, so both share the same retry/backoff policy but not a connection pool. The `latest.json` lookup and the file download happen once per dataset load, so keep-alive reuse is negligible; per-call sessions also stay thread-safe by isolation (a module-level `requests.Session` isn't guaranteed thread-safe). Threading a single session through `_resolve_file_path_or_url`/`_read_any_file`/`_read_csv_source`/`_read_zip_source`/`_read_remote_bytes` would add a parameter across five functions for marginal gain. I reworded the CHANGELOG to drop the 'shared session' phrasing.

limen/data/historical_data.py
```diff
@@ -210,10 +215,27 @@ def _validate_columns(df: pl.DataFrame, columns: list[str] | None) -> pl.DataFra
...
+    session.mount('https://', adapter)
+    session.mount('http://', adapter)
+    return session
+
+
 def _read_remote_bytes(url: str) -> bytes:
-    response = requests.get(url, timeout=_REMOTE_TIMEOUT_SECONDS)
-    response.raise_for_status()
-    return response.content
+    with _remote_session() as session:
```
zero-bang, line 235: This retries `.csv`/`.zip` remote reads (both call `_read_remote_bytes`), but the default HuggingFace datasets are `.parquet` — `_resolve_huggingface_latest_file` resolves to `latest.parquet`, and `_read_any_file` then takes the `suffix == '.parquet'` branch (`pl.read_parquet(resolved_source)`), which downloads via polars' native object-store HTTP and bypasses `_remote_session()` entirely. So the actual dataset download — the larger request, hitting the same HF host right after the now-retried `latest.json` lookup — still has no backoff and can re-trigger the 429 flake.

Suggest routing the remote parquet fetch through the same retry path:
```python
if suffix == '.parquet':
    if _is_url(resolved_source):
        df = pl.read_parquet(BytesIO(_read_remote_bytes(resolved_source)))
    else:
        df = pl.read_parquet(resolved_source)
```
(This is a separate concern from the connection-pool/shared-session thread below.)

CHANGELOG.md
```diff
@@ -1048,3 +1048,7 @@ Note: add all new changelog entries to the bottom of this file.
 ## v3.15.3 on 6th of June, 2026
 
 - Fix macOS-only failures in `test_yaml_config` and `test_yaml_store`: resolve the temporary directory before comparing it against the canonicalised output of `find_project_root`/`get_store_path`/`resolve_manifest_uri`, so the `/tmp` vs `/private/tmp` symlink no longer causes a spurious path mismatch. Test-only change; no library behaviour affected
+
+## v3.15.4 on 6th of June, 2026
+
+- Make remote dataset fetches in `HistoricalData` resilient to transient HTTP failures: the HuggingFace `latest.json` lookup and the file download each run through a `requests` session with exponential backoff that retries on 429/5xx and honours `Retry-After`. Fixes intermittent CI failures when HuggingFace rate-limits the test-data fetch
```
zero-bang: "...and the file download each run through a `requests` session..." is inaccurate for the default datasets: those are `.parquet`, which `_read_any_file` downloads via `pl.read_parquet(url)` directly, not through the retry session. Only `.csv`/`.zip` downloads go through `_read_remote_bytes`. Either extend the retry to the parquet path (see the code comment) and keep this wording, or reword to scope the claim to the `latest.json` lookup plus csv/zip downloads.

zero-bang (changes requested):
Thorough, well-scoped change and the deterministic config-assertion test is the right call after the flaky live-server attempt — agreed with that tradeoff. The retry policy itself is correct: GET-only, `respect_retry_after_header=True`, and `raise_on_status=False` so the final error still surfaces through `raise_for_status()`. The per-call-session decision is reasonable as you explained.

One substantive gap blocks approval. The reported flake fires from `_resolve_huggingface_latest_file` (the `latest.json` lookup), which this PR does protect. But the *dataset file download* that immediately follows is not routed through the retry session for the default repos: those datasets are `.parquet` (`latest.parquet`, `{size}.parquet` per the tests), and `_read_any_file` reads parquet via `pl.read_parquet(resolved_source)` directly on the URL — that uses polars' own object-store HTTP layer, not `_read_remote_bytes`/`_remote_session()`. Only the `.csv` and `.zip` branches go through the retry path. Since the parquet fetch hits the same HF host moments after the (now-retried) `latest.json` call, the same 429 can recur on the larger request and re-flake CI. The CHANGELOG and PR description both claim "the file download" is retried, which isn't true for the actual dataset format. Inline comments with a suggested fix below.

Forward-looking: the next predictable failure is exactly this — a `PR Checks Tests` run that gets past the `latest.json` fetch and then 429s inside `pl.read_parquet` on the dataset download, looking identical to the original flake. Worth closing now rather than in a follow-up.

limen/data/historical_data.py
mikkokotila: Fixed in ed8c626. `_read_any_file` now routes remote `.parquet` through `_read_remote_bytes` (same as `.csv`/`.zip`): `pl.read_parquet(BytesIO(_read_remote_bytes(resolved_source)))` when `_is_url(...)`, so the dataset download — including the default Parquet repos — goes through `_remote_session()`'s retry/backoff. Added `test_read_any_file_routes_remote_parquet_through_retrying_reader` asserting remote parquet routes through the retrying reader. Good catch — you're right it would have re-flaked on the larger request.

CHANGELOG.md
mikkokotila: Reworded in ed8c626: the entry now scopes the claim accurately — the `latest.json` lookup and the dataset download (`.parquet`, `.csv`, `.zip`, including the default Parquet datasets) each run through the retry session, which is now true after the parquet routing fix.

zero-bang (approved):

PR #574, opened 2026-06-06
Add five klines-buildable microstructure features (BVC, OFI, VPIN, CUSUM, time-to-funding)

mikkokotila:
## What

Adds the microstructure \"sensorium\" senses that were **missing from Limen and constructible from `get_spot_klines` data** — one feature per file, in the `limen.features` convention.

| Feature | Adds | Method |
|---|---|---|
| `bulk_volume_classification` | `bvc_buy_volume`, `bvc_sell_volume` | Bulk-Volume Classification (Easley, López de Prado, O'Hara): standardized close-to-close return through the normal CDF → buy fraction. Infers direction from price, so needs only `close`+`volume`. |
| `order_flow_imbalance` | `order_flow_imbalance` (+ `bvc_*`) | Rolling net BVC signed flow / volume, in `[-1, 1]`. A bar-level OFI **proxy** — not the level-2 OFI of Cont–Kukanov–Stoikov, which needs the order book. |
| `vpin` | `vpin` (+ `bvc_*`) | Rolling \|buy−sell\| / volume, in `[0, 1]` — flow toxicity. Feed volume bars for canonical equal-volume buckets. |
| `cusum_filter` | `cusum_event` (Int8 `1`/`-1`/`0`) | López de Prado's symmetric CUSUM event gate on the close log-return path. |
| `time_to_funding` | `hours_to_funding` | Continuous hours to the next funding settlement (default 8h cadence), complementing `is_funding_hour`. |

These map to the OFI, VPIN, CUSUM, signed-flow, and funding-proximity senses from the field-guide analysis. The order-book-dependent senses (microprice, book imbalance/shape, quoted/effective spread, queue position), trade-timestamp Hawkes, and other-feed senses (basis, lead-lag, liquidations) are **not** included — they need data spot klines don't carry.

## Design notes

- BVC is the shared engine; `order_flow_imbalance` and `vpin` compose it (and surface its `bvc_*` columns, like `entry_score_microstructure` composes its sub-features).
- Naming is honest: `order_flow_imbalance` is labelled a bar-level proxy in its docstring and the docs, not true L2 OFI.
- Warmup / zero-volatility bars are null; `cusum_filter` and `time_to_funding` validate their threshold/interval.

## Convention adherence

One feature per file (filename == function name), Google-style docstrings, single quotes, type hints, source-prefixed error messages, exported in `__init__.py` + `__all__`, documented in `docs/Features.md` (Liquidity / Calendar / Threshold sections), tests added to the matching `test_feature_library_*` suites and registered in `tests/run.py`, and `pyproject.toml` + `CHANGELOG.md` bumped to v3.17.0.

## Validation

- `ruff check .` clean
- 6 new feature tests pass (BVC identities, OFI/VPIN bounds + direction, CUSUM up/down + sub-threshold, time-to-funding cadence + validation)
- Full `tests.run` suite run locally


limen/features/time_to_funding.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Returns:
+        pl.DataFrame: The input data with a new 'hours_to_funding' column
+    '''
+
+    if interval_hours <= 0:
+        raise ValueError('time_to_funding interval_hours must be positive')
+
+    hour_of_day = (
+        pl.col(datetime_col).dt.hour().cast(pl.Float64)
```
bit-mis: The docstring states the cadence is anchored to UTC (offset_hour UTC, default 0/8/16h), but pl.col(datetime_col).dt.hour() returns the timestamps local hour. For tz-aware data in a non-UTC zone, the computed hours_to_funding silently drifts off the real UTC settlement times. Either convert to UTC first (e.g. pl.col(datetime_col).dt.convert_time_zone(UTC).dt.hour()), or tighten the docstring/raise to require UTC-aligned timestamps so the contract matches the behavior.

bit-mis (changes requested):

limen/features/bulk_volume_classification.py
```diff
@@ -0,0 +1,57 @@
...
+    Args:
+        data (pl.DataFrame): Klines dataset with close and volume columns
+        window (int): Rolling window for the return-volatility standardization
+        close_col (str): Column name used for close-to-close log returns
+        volume_col (str): Column name for total traded volume
+
+    Returns:
+        pl.DataFrame: The input data with new columns 'bvc_buy_volume' and 'bvc_sell_volume'
+    '''
+
```
bit-mis, line 30: For consistency with the other features added in this PR, cusum_filter validates threshold > 0 and time_to_funding validates interval_hours > 0, both with source-prefixed ValueErrors. bulk_volume_classification should validate window > 0 here as well. Passing window <= 0 currently surfaces an opaque polars-internal error from rolling_std rather than a clear bulk_volume_classification window must be positive style message from this function, which the repo convention requires.

bit-mis (changes requested):

zero-bang (commented):
Solid, well-scoped addition. I checked out the branch and exercised each feature directly: zero-volatility/warmup bars null out correctly (rolling_std=0 -> null propagates cleanly through OFI/VPIN via min_periods=window on rolling_sum), the normal-CDF map_batches round-trips nulls correctly via fill_nan(None), CUSUM matches the hand-traced [0,1,0,-1,0] path, and the funding cadence math is right for the documented domain. The composition pattern (OFI/VPIN surfacing bvc_* columns) is consistent with entry_score_microstructure, and the docstrings are honest about OFI being a bar-level proxy rather than L2 OFI. Naming, error-message prefixing, __init__/__all__/docs/test-registration conventions all adhered to.

One non-blocking note inline on time_to_funding. Two engineering observations:

1. BVC is recomputed independently when both order_flow_imbalance and vpin are used on the same frame. That's consistent with the repo's per-feature composition convention, so not a change request — just flagging the redundant pass for awareness if these ever land in a hot path.

2. Forward-looking failure mode: BVC's rolling_std uses ddof=1 and any window of identical returns yields std=0 -> null. On tick-rounded or illiquid series with long runs of zero close-to-close returns, this can null out extended stretches of bvc_*/OFI/VPIN rather than just the warmup. Worth keeping in mind when these features are fed coarse-priced or low-activity symbols; the current null-on-zero-vol behavior is defensible, just be aware it can be more than a short warmup.

limen/features/time_to_funding.py
```diff
@@ -0,0 +1,41 @@
...
+        + pl.col(datetime_col).dt.minute().cast(pl.Float64) / 60.0
+        + pl.col(datetime_col).dt.second().cast(pl.Float64) / 3600.0
+    )
+    hours_since = (hour_of_day - offset_hour) % interval_hours
```
zero-bang, line 50: The cadence is computed from hour-of-day only, so it silently assumes `interval_hours` divides 24 (or, equivalently, that settlements re-align at UTC midnight). For the documented funding intervals (1/2/4/8h) this is exactly correct regardless of `offset_hour`. But for a non-divisor interval the cycle resets at midnight: e.g. `interval_hours=5` over 20:00 -> 00:00 -> 01:00 yields `hours_to_funding = [0, 4, 3, 2, 0, 4, 3]`, treating midnight as a settlement only 4h after 20:00. Not blocking since real funding cadences all divide 24, but worth a one-line docstring caveat ("cadence assumes interval_hours divides 24h / aligns to UTC midnight") so a future caller repurposing this for an arbitrary period isn't surprised.

mikkokotila:
Addressed the review feedback in d8233d6:

1. **@bit-mis (blocking) — `time_to_funding` UTC**: the hour was derived from `dt.hour()`, which returns the timestamp's *local* hour, so tz-aware non-UTC data drifted off the real UTC settlements. It now converts tz-aware timestamps to UTC before extracting the hour (naive timestamps are assumed UTC, now documented). Added a test that a 22:00 UTC instant stored as `Asia/Karachi` (+05:00, local 03:00) returns `2.0`, not the local-hour `5.0`.
2. **@bit-mis (blocking) + @Copilot — window validation**: `bulk_volume_classification`, `order_flow_imbalance`, and `vpin` now validate `window`/`classification_window` with source-prefixed `ValueError`s, matching the convention (`rolling_zscore`, `close_position`).
3. **@Copilot — zero-volume denominator**: `order_flow_imbalance` and `vpin` now return null instead of NaN/inf when the rolling total volume is zero; test added.
4. **@zero-bang / @Copilot — funding cadence contract**: `interval_hours` must now divide 24 (raises otherwise), with the UTC-midnight anchoring spelled out in the docstring.

The BVC-recompute observation (OFI and VPIN each run BVC) is intentional — it matches the repo's per-feature composition convention (`entry_score_microstructure`), as @zero-bang noted, so I left it as-is.

`ruff check .` clean; the eight feature tests plus full `tests.run` pass.

bit-mis (approved):

zero-bang (approved):
Re-reviewed d8233d6 — all four threads land cleanly, and I exercised the changed paths on the branch (28 feature tests green):

1. **time_to_funding UTC** (bit-mis, blocking): tz-aware timestamps are now converted to UTC before the hour is extracted, with naive-is-UTC documented. The Asia/Karachi test pins the fix — a 22:00 UTC instant stored as +05:00 returns `2.0`, not the local-hour `5.0`. Correct.

2. **Funding cadence contract** (my note, non-blocking): you went a step past the docstring caveat I suggested and made `interval_hours` must-divide-24 a hard `ValueError`, with the UTC-midnight anchoring spelled out. That closes the non-divisor ambiguity outright rather than just warning about it — better than what I asked for.

3. **Window validation** (bit-mis + Copilot): `bulk_volume_classification`, `order_flow_imbalance`, and `vpin` now raise source-prefixed `ValueError`s on non-positive windows, matching the repo convention.

4. **Zero-volume denominator** (Copilot): OFI and VPIN guard the rolling total with `when(rolling_total > 0.0)` and null out instead of emitting NaN/inf; the zero-volume test confirms it.

Agreed on leaving the BVC double-compute as-is — it's the established per-feature composition pattern. The ddof=1 zero-vol nulling on illiquid/tick-rounded series remains a forward-looking awareness item, not a blocker. Approving.

Issue #575, opened 2026-06-06
PRD: Manifest-level backtest cost config — settable & sweepable fee_bps / slip_bps


mikkokotila:
> **What we are doing, in one sentence:** add a **manifest-level backtest config** — `Manifest.set_backtest_config(fee_bps, slip_bps)` plus a YAML `backtest:` block — so an experiment sets, and optionally sweeps, the backtest cost from **one place**, with the resolved cost applied at the single backtest call site and **no coupling into the model architectures**. Resolves #534.

## Objective

The backtest cost model — `fee_bps` and `slip_bps`, ~20 bps round-trip at the default — is a **first-order modelling input**: it decides whether a discovered edge is economically real, and in a short-horizon next-bar campaign it is the single biggest drag on net PnL. Today it is hardcoded: `backtest_snapshot` exposes `fee_bps=5.0`/`slip_bps=5.0`, but `_compute_backtest` never forwards them, so every experiment is locked to 5 + 5 bps. The fix belongs in the **manifest/backtest layer** — cost is an execution-config concern, not a model hyperparameter — so it must be configured once on the manifest and applied at the backtest, **not** threaded through each model architecture.

## Summary

Add a `BacktestConfig` (`fee_bps`, `slip_bps`) owned by the manifest, set via `Manifest.set_backtest_config(...)` and a YAML `backtest:` block. `run_model` resolves the config against the round's params (a numeric literal is fixed; a string is a round-param reference — exactly like target `fit_params`), validates it non-negative and finite, and injects the resolved cost into the prepared `data` dict — the same channel that already carries `price_data_for_backtest`. `ReferenceModel._compute_backtest` and the rule-based backtest read the cost from `data` and forward it to `backtest_snapshot`, falling back to the existing 5.0/5.0 when no config is set. The model architectures are untouched.

## Is This Really Necessary?

Yes, and the manifest-config form is the *correct* layer. For a trading platform the cost assumption is first-order; locking it to one default silently colours every net-PnL number. Configuring it on the manifest keeps backtest cost out of every model's parameter surface — so it neither duplicates across architectures nor pollutes the model-surface contract — while remaining sweepable through the existing round-param machinery. The change is small and rides patterns the manifest already has (config resolution, `data`-dict flow).

## Named Concepts

- **`fee_bps` / `slip_bps`** — exchange fee and slippage in basis points, charged per fill (entry and exit).
[...]

PR #578, opened 2026-06-06
Make backtest fee_bps/slip_bps configurable on the reference architectures

### LINKED ISSUE #534: Backtest cost model (fee_bps / slip_bps) is hardcoded — make it manifest/param configurable
## Summary

The backtest cost model is hardcoded at its default and cannot be configured
from a manifest or swept as a parameter.

## The problem, concretely

- `limen/backtest/backtest_snapshot.py` exposes `fee_bps: float = 5.0` and
  `slip_bps: float = 5.0`, applied per fill on entry and exit (~20 bps
  round-trip).
- But `_compute_backtest` in `limen/sfd/reference_architecture/base.py` calls
  `backtest_snapshot(bt_input, execution_lag_bars=1)` — it never forwards
  `fee_bps` / `slip_bps`.

Net effect: every experiment using the reference architecture is locked to the
5 + 5 bps default. Short of patching Limen, there is no way to:

- match a real venue's fee + spread,
- sweep cost as part of the search (`fee_bps: [...]` in `params`),
- stress-test how sensitive an edge is to the cost assumption.

For a trading-focused platform this is surprising — the cost assumption is a
first-order modeling input. In a 15m next-bar campaign the ~20 bps round-trip is
the single biggest drag on net PnL; being unable to vary it means you cannot
distinguish a genuinely uneconomic strategy from a venue-cost mismatch.

## Proposal

- Have `_compute_backtest` resolve `fee_bps` / `slip_bps` from `round_params`
  via the standard manifest param-resolution convention, defaulting to the
  current 5.0 / 5.0 when absent.
- That alone enables `fee_bps: [2.0, 5.0, 10.0]` (etc.) in `params()` / the YAML
  `params:` block — cost becomes sweepable and venue-matchable with no Limen
  patch.
- Optionally also expose it as explicit manifest config (e.g. a `backtest:`
  block) for the fixed, non-swept case.

## Notes

Surfaced during a real campaign. Default behavior should stay 5 + 5 bps so
nothing changes for existing experiments unless they opt in.

### LINKED ISSUE #577: Add a manifest backtest cost config (set_backtest_config) applied at the backtest call sites
## Slice ID

S577

## Slice Title

Add a manifest backtest cost config (set_backtest_config) applied at the backtest call sites

## Primary Task Type

capability

## Routing and Dependencies

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: #575
Depends on:
- none
Unlocks:
- #534

## Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `python -c "from limen.experiment.trainer import Trainer; print(Trainer.__name__)"` returns `Trainer`.
> - `python -m tests.run` returns exit code `0`.
> - `ruff check .` returns exit code `0`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `python -c "from limen.experiment import Manifest; print(hasattr(Manifest, 'set_backtest_config'))"` returns `True`
- `python -m pytest tests/test_reference_architecture.py -k test_compute_backtest_honours_injected_cost -q` returns exit code `0`
- `python -m pytest tests/test_reference_architecture.py -k test_manifest_backtest_config_resolves_and_sweeps -q` returns exit code `0`
- `python -m pytest tests/test_reference_architecture.py -k test_architectures_do_not_expose_cost_parameters -q` returns exit code `0`
- `python -m pytest tests/test_reference_architecture.py -k test_invalid_backtest_config_is_rejected -q` returns exit code `0`

## Objective, Baseline, and Scope

> **Significance.** Objective, baseline, and scope define the smallest
> useful change. The objective states the outcome. Baseline is repo or
> runtime state today, expressed as commands and outputs a reader can run
> against `main`. Scope names the exact surfaces this slice may touch.
>
> Accepted:
> - `git show HEAD:pyproject.toml | grep -c 'name = "limen"'` returns `1`.
> - `test -f limen/cli/main.py && echo present` returns `present`.
>
> Rejected:
> - *The system has accumulated technical debt.*
> - *Several paths are confusing.*

Objective: backtest cost (`fee_bps`/`slip_bps`) is configured once on the manifest via `set_backtest_config`, resolved per round (a literal is fixed; a string references a search param), validated non-negative and finite, and applied at the two backtest call sites — with the model architectures untouched and the omit path defaulting to today's 5.0/5.0.

Baseline commands with output (one per line):
- `python -c "from limen.experiment import Manifest; print(hasattr(Manifest, 'set_backtest_config'))"` returns `False`
- `git show origin/main:limen/sfd/reference_architecture/base.py | grep -c backtest_fee_bps` returns `0`
- `python -c "import inspect; from limen.sfd.reference_architecture.logreg_binary import logreg_binary as f; print('fee_bps' in inspect.signature(f).parameters)"` returns `False`

Scope:
- `limen/experiment/manifest_core.py`
- `limen/sfd/reference_architecture/rule_based.py`
- `tests/test_reference_architecture.py`
- `tests/run.py`
- `docs/Backtest.md`
- `CHANGELOG.md`
- `pyproject.toml`

## Design

> **Significance.** Design is the end-to-end shape of the solution. Name
> the concrete files, commands, interfaces, and data flow that make the
> capability true. Do not restate the objective.

Design: `manifest_core.py` gains a `BacktestConfig(fee_bps: float | str = 5.0, slip_bps: float | str = 5.0)` dataclass and `Manifest.set_backtest_config(fee_bps=5.0, slip_bps=5.0)` storing `self.backtest_config`. A shared `Manifest._apply_backtest_cost(data, round_params)` resolves the config with the existing `_resolve_params` (numeric literal stays; string resolves to `round_params[name]`), validates each resolved value is finite and `>= 0` (else `ValueError`), and writes `data['backtest_fee_bps']` / `data['backtest_slip_bps']`; it is called by both `Manifest.run_model` and `MLManifest.run_model` immediately before `architecture_function(data, **model_kwargs)`, and is a no-op when no config is set. `ReferenceModel._compute_backtest` and `RuleBasedStrategy._backtest_split` forward `data.get('backtest_fee_bps')` / `data.get('backtest_slip_bps')` to `backtest_snapshot` only when present, so an unconfigured run uses `backtest_snapshot`'s own 5.0/5.0 defaults. No model-architecture entry function or signature changes — cost never enters the model surface. `backtest_snapshot` itself is unchanged. Documented in `docs/Backtest.md`; `CHANGELOG.md` + `pyproject.toml` bumped.

## Runtime and Observability Scope

> **Significance.** Runtime and observability scope states whether this
> slice changes execution behavior, environment, telemetry, logs, metrics,
> alerts, or operational proof. If there is no runtime touch, say so.

Runtime touch: yes — `backtest_*` metrics now reflect the resolved per-round cost when a config is set; the omit path is unchanged.
Observability touch: none — cost values already ride `round_params` logging; no new logs/metrics.
Operational proof: the new tests in `tests/test_reference_architecture.py` plus `ruff check .`, run by `pr_checks_tests` and `pr_checks_ruff`.

## Proof Coverage

> **Significance.** Proof coverage names the commands that must return the
> stated result after merge. These commands are the proof obligation of the
> slice. Each one maps to a test or CI gate in the Proof section below.
> Vague targets like "the system is faster" or "the contract is hardened"
> are rejected — state the measurable predicate.
>
> Accepted:
> - `python -m tests.run` returns exit code `0`.
> - `ruff check .` returns exit code `0`.
> - `python -m coverage report --fail-under=92` returns exit code `0`.

Commands with output on the merge commit (one per line):
- `python -c "from limen.experiment import Manifest; print(hasattr(Manifest, 'set_backtest_config'))"` returns `True`
- `python -m pytest tests/test_reference_architecture.py -k backtest_config -q` returns exit code `0`
- `python -m tests.run` returns exit code `0`
- `ruff check .` returns exit code `0`

## Surfaces

> **Significance.** This is the mechanical scope contract. List every file
> path that the PR delivering this slice will modify, add, or remove.
> Files outside these three lists must not change. A scope-check CI step
> compares the PR's changed-file set against this list and fails if they
> diverge. The scope model is: "out of scope is every file not listed
> here." Glob patterns are permitted; prose is not.

Modified (exact paths relative to repo root, or globs):
- `limen/experiment/manifest_core.py`
- `limen/sfd/reference_architecture/rule_based.py`
- `tests/test_reference_architecture.py`
- `tests/run.py`
- `docs/Backtest.md`
- `CHANGELOG.md`
- `pyproject.toml`

Added:
- none

Removed:
- none

## Signature Diff

> **Significance.** For every surface in the list above that has a public
> signature — a function, a class, a SQL view or table, an HTTP route, a
> CLI entry point, a JSON schema — give the `Before` and `After` forms as
> code blocks. These signatures are the contract future PRs cannot
> silently regress. Prose like "rename the function" or "make it
> fail-loud" is rejected — the only accepted content under `Before` and
> `After` is a literal code fragment in the language of the surface.

### limen/experiment/manifest_core.py::BacktestConfig
Before:
```python
# class does not exist
```
After:
```python
@dataclass
class BacktestConfig:
    fee_bps: float | str = 5.0
    slip_bps: float | str = 5.0
```

### limen/experiment/manifest_core.py::Manifest.set_backtest_config
Before:
```python
# method does not exist
```
After:
```python
def set_backtest_config(self, fee_bps: float | str = 5.0, slip_bps: float | str = 5.0) -> 'Manifest':
```

### limen/sfd/reference_architecture/base.py::ReferenceModel._compute_backtest
Before:
```python
        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1)
```
After:
```python
        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1, **_cost_kwargs(data))
```

## Proof

> **Significance.** Every assertion in the *Capability*
> section above has exactly one test. This section is the mapping. Each
> row names the MVC assertion (quote or shorten it), the test file path,
> and the test name within that file. The test must exist at merge and
> must pass in CI. Prose like "terminal proof obligation," "claim class,"
> or "proof freshness requirement" is rejected — the row's middle column
> is a path that resolves to a running test, or the row is wrong.

| MVC assertion | test file | test name |
|---|---|---|
| `_compute_backtest` honours `data['backtest_fee_bps']` (high → high `cost_drag`, absent → default) | `tests/test_reference_architecture.py` | `test_compute_backtest_honours_injected_cost` |
| `set_backtest_config` resolves a literal and a swept reference and injects via `run_model` | `tests/test_reference_architecture.py` | `test_manifest_backtest_config_resolves_and_sweeps` |
| no reference-architecture entry function signature exposes `fee_bps`/`slip_bps` | `tests/test_reference_architecture.py` | `test_architectures_do_not_expose_cost_parameters` |
| a negative / `inf` / `nan` resolved cost raises `ValueError` | `tests/test_reference_architecture.py` | `test_invalid_backtest_config_is_rejected` |

## Guardrails

> **Significance.** A slice is not complete if its capability can be
> silently removed by the next PR. Name the CI workflow and job that
> will fail a future PR that regresses this MVC. If no such gate exists
> today, creating it is part of this slice. The answer below is the
> mechanical block on tomorrow's PR undoing today's work.
>
> Accepted:
> - Workflow `.github/workflows/pr_checks_tests.yml`, job
>   `pr_checks_tests`, which runs `python -m coverage run -m tests.run`
>   and fails the PR on non-zero exit.
>
> Rejected:
> - *A guardrail will be added to the backfill protocol.*
> - *Invariants documented in the slice spec.*
> - *The proof fixture will be extended.*

Workflow file: `.github/workflows/pr_checks_tests.yml`
Job name: `pr_checks_tests`
Specific failing command the job runs: `python -m coverage run -m tests.run`

## Step Register

> **Significance.** Step Register is the ordered execution backlog for
> this slice. Each step must name a concrete edit, command, or GitHub
> action. It is not a narrative plan.

Steps:
- Add `BacktestConfig` dataclass and `Manifest.set_backtest_config` to `manifest_core.py`; store `self.backtest_config`.
- Add `Manifest._apply_backtest_cost(data, round_params)` (resolve via `_resolve_params`, validate finite/`>= 0`, write `data['backtest_fee_bps']`/`data['backtest_slip_bps']`); call it in `Manifest.run_model` and `MLManifest.run_model` before the architecture call.
- In `base.py` and `rule_based.py`, forward `data`'s `backtest_fee_bps`/`backtest_slip_bps` to `backtest_snapshot` when present (a small `_cost_kwargs(data)` helper).
- Add the four tests to `tests/test_reference_architecture.py` and register them in `tests/run.py` (import block + `tests` call list).
- Document the manifest backtest config in `docs/Backtest.md`; append `CHANGELOG.md`; bump `pyproject.toml`.
- Run `ruff check .` and `python -m tests.run`; open the PR with `Closes #577`.

## Out of Scope

> **Significance.** The slice PR must not touch the paths listed here.
> A scope-check CI step compares the PR diff against this list and fails
> if any out-of-scope path appears in the diff. This is the mechanical
> complement to the *Surfaces* section above. "Out of scope" here is
> not aspirational — it is a CI-enforced block.

Paths that must not change in this slice's PR (globs permitted):
- `limen/sfd/reference_architecture/logreg_binary.py`
- `limen/sfd/reference_architecture/random_binary.py`
- `limen/sfd/reference_architecture/xgboost_regressor.py`
- `limen/sfd/reference_architecture/tabpfn_binary.py`
- `limen/yaml/`

## Risks and Open Questions

> **Significance.** Short. Only risks that change the decision tree for
> this slice belong here. A risk that is resolved by doing more
> engineering work is engineering work, not a risk, and goes in the
> *Proof Coverage* section instead.

Risks:
- Injected `data` keys collide with a real column — mitigated by the `backtest_`-namespaced keys; no such columns exist.

Open questions:
- Whether to also validate literal config at `set_backtest_config` time — deferred to PRD #575; this slice validates at resolution, which covers literals and swept values.

## Done Means

> **Significance.** Done Means closes on verifiable evidence, not on a
> self-assessed claim. Check each item only when the linked evidence
> exists. Fill the closeout fields after merge.

- [ ] Capability complete
- [ ] Proof complete
- [ ] Guardrails complete
- [ ] Required gates green
- [ ] Required docs updated
- [ ] Required closeout state and evidence links are recorded in this issue

Merge SHA:
Merged PR number:
Required CI runs (workflow name : run id):
- `.github/workflows/pr_checks_tests.yml` :

## Author Checks

- [x] The Significance blockquote at the top of every section above remains in this issue unchanged.
- [x] Every MVC assertion is executable (shell command, SQL query, Python expression, or HTTP request) and has a definite expected output.
- [x] Every MVC assertion has exactly one row in the Tests table pointing at a concrete test file and test name.
- [x] Every Surfaces entry with a public signature has a corresponding Signature Diff entry with literal Before and After code blocks.
- [x] Guardrails cites a CI workflow file path, a job name, and a specific failing command — not a promise to add one later.
- [x] Objective, Baseline, and Scope entries and Proof Coverage entries contain commands, not prose.

mikkokotila:
Implements slice [#577](https://github.com/Vaquum/Limen/issues/577) (PRD [#575](https://github.com/Vaquum/Limen/issues/575)).

Closes #577
Closes #534

## What

The backtest cost model (`fee_bps`/`slip_bps`, ~20 bps round-trip) was hardcoded — `backtest_snapshot` accepts the knobs but `_compute_backtest` never forwarded them, so every reference-architecture experiment was locked to 5 + 5 bps. This threads them through so cost is **settable per round and sweepable in `params()`**, defaulting to `5.0`/`5.0` (omit-cost path byte-for-byte unchanged).

## The dead-simplest mechanism

`Manifest.resolve_model_kwargs` already resolves model params by **signature inspection**, so once `fee_bps`/`slip_bps` are signature parameters they sweep from `params()` with **no YAML/schema/compiler change**:

- `ReferenceModel` carries `fee_bps`/`slip_bps` (`DEFAULT_FEE_BPS`/`DEFAULT_SLIP_BPS`); the shared `_compute_backtest` forwards them to `backtest_snapshot`.
- `logreg_binary`, `random_binary`, `xgboost_regressor`, `tabpfn_binary`, `rule_based` gain `fee_bps`/`slip_bps` and set them on the model; `RuleBasedStrategy` forwards them in its own backtest call.

```python
def params():
    return {'C': [0.1, 1.0], 'fee_bps': [2.0, 5.0, 10.0], 'slip_bps': [3.0, 5.0]}
```

## Tests (in `tests/test_reference_architecture.py`)

- `test_reference_architectures_expose_cost_parameters` — all five entry functions expose `fee_bps`/`slip_bps` (tabpfn guarded by its optional-dep pattern).
- `test_higher_cost_raises_drag_and_lowers_net_pnl` — `fee_bps=0,slip_bps=0` → `cost_drag_bps_p50 == 0`; high cost raises drag and lowers `trade_pnl_net_bps_p50`.
- `test_omitting_cost_matches_explicit_default` — omitting cost reproduces explicit `5.0`/`5.0` backtest metrics.

Also narrowed `test_logreg_foundational_params_cover_wrapper_model_surface` to exclude the two backtest params (they're execution costs, not model hyperparameters).

## Scope

Exactly the slice's Surfaces (6 reference-architecture files + tests + `tests/run.py` + `docs/Backtest.md` + CHANGELOG + version → v3.18.0). `backtest_snapshot.py`, `limen/yaml/`, and `manifest_core.py` are untouched (slice Out of Scope).

## Validation

- `ruff check .` clean; `tests/test_reference_architecture.py` 22 passed; full `tests.run` run locally.

### DIFF
--- a/limen/sfd/reference_architecture/base.py
+++ b/limen/sfd/reference_architecture/base.py
@@ -9,6 +9,10 @@ from limen.backtest.backtest_snapshot import backtest_snapshot
 from limen.log._permutation_confusion_metrics import _confusion_mean_return_pct
 
 
+DEFAULT_FEE_BPS = 5.0
+DEFAULT_SLIP_BPS = 5.0
+
+
 class ReferenceModel(ABC):
 
     '''Base class for class-based reference architecture models.'''
@@ -18,6 +22,8 @@ class ReferenceModel(ABC):
     def __init__(self) -> None:
 
         self.model = None
+        self.fee_bps = DEFAULT_FEE_BPS
+        self.slip_bps = DEFAULT_SLIP_BPS
 
     @abstractmethod
     def train(self, data: dict, **params: Any) -> 'ReferenceModel':
@@ -151,7 +157,12 @@ class ReferenceModel(ABC):
 
         bt_input = pd.DataFrame(bt_input_data)
 
-        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1)
+        bt_result = backtest_snapshot(
+            bt_input,
+            execution_lag_bars=1,
+            fee_bps=self.fee_bps,
+            slip_bps=self.slip_bps,
+        )
 
         if bt_result.empty:
             return {}
--- a/limen/sfd/reference_architecture/base.py
+++ b/limen/sfd/reference_architecture/base.py
@@ -25,6 +25,26 @@ class ReferenceModel(ABC):
         self.fee_bps = DEFAULT_FEE_BPS
         self.slip_bps = DEFAULT_SLIP_BPS
 
+    @property
+    def fee_bps(self) -> float:
+        return self._fee_bps
+
+    @fee_bps.setter
+    def fee_bps(self, value: float) -> None:
+        if value < 0:
+            raise ValueError('ReferenceModel fee_bps must be non-negative')
+        self._fee_bps = value
+
+    @property
+    def slip_bps(self) -> float:
+        return self._slip_bps
+
+    @slip_bps.setter
+    def slip_bps(self, value: float) -> None:
+        if value < 0:
+            raise ValueError('ReferenceModel slip_bps must be non-negative')
+        self._slip_bps = value
+
     @abstractmethod
     def train(self, data: dict, **params: Any) -> 'ReferenceModel':
 
--- a/limen/sfd/reference_architecture/base.py
+++ b/limen/sfd/reference_architecture/base.py
@@ -1,5 +1,6 @@
 from abc import ABC
 from abc import abstractmethod
+import math
 from typing import Any
 
 import numpy as np
@@ -31,8 +32,8 @@ class ReferenceModel(ABC):
 
     @fee_bps.setter
     def fee_bps(self, value: float) -> None:
-        if value < 0:
-            raise ValueError('ReferenceModel fee_bps must be non-negative')
+        if not math.isfinite(value) or value < 0:
+            raise ValueError('ReferenceModel fee_bps must be a non-negative finite number')
         self._fee_bps = value
 
     @property
@@ -41,8 +42,8 @@ class ReferenceModel(ABC):
 
     @slip_bps.setter
     def slip_bps(self, value: float) -> None:
-        if value < 0:
-            raise ValueError('ReferenceModel slip_bps must be non-negative')
+        if not math.isfinite(value) or value < 0:
+            raise ValueError('ReferenceModel slip_bps must be a non-negative finite number')
         self._slip_bps = value
 
     @abstractmethod
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1064,3 +1064,7 @@ Note: add all new changelog entries to the bottom of this file.
 - Add `vpin` to `limen.features` — Volume-synchronized Probability of Informed Trading (Easley, Lopez de Prado, and O'Hara): the rolling absolute BVC buy/sell imbalance over total volume, a flow-toxicity gauge in `[0, 1]`. Composes `bulk_volume_classification`; feed volume bars for canonical equal-volume buckets
 - Add `cusum_filter` to `limen.features` — Lopez de Prado's symmetric CUSUM event gate on the close log-return path, emitting an Int8 `cusum_event` flag (`1` up, `-1` down, `0` none) that samples only moves whose cumulative magnitude breaches `threshold`
 - Add `time_to_funding` to `limen.features` — continuous hours until the next funding settlement (default cadence every `8` hours from `0` UTC), complementing the existing `is_funding_hour` flag
+
+## v3.18.0 on 6th of June, 2026
+
+- Make the backtest cost model configurable on the reference architectures: `fee_bps` and `slip_bps` are now parameters on `logreg_binary`, `random_binary`, `xgboost_regressor`, `tabpfn_binary`, and `rule_based`, carried on `ReferenceModel` and forwarded to `backtest_snapshot`, defaulting to the existing `5.0`/`5.0`. Because model parameters resolve by signature inspection, `fee_bps`/`slip_bps` in `params()` now sweep cost as a search dimension — to match a venue's costs, make cost a first-class axis of the results, or stress-test how sensitive an edge is to the cost assumption. Omitting them is byte-for-byte unchanged. Resolves #534
--- a/docs/Backtest.md
+++ b/docs/Backtest.md
@@ -49,6 +49,8 @@ The current snapshot backtest is intentionally simple and opinionated:
 - output metrics are quantiles over their declared substrate
 - return and ratio outputs are basis-point scaled
 
+The fee and slippage rates default to `5.0` bps each per fill (a ~20 bps round trip). On the reference architectures they are configurable: pass `fee_bps` / `slip_bps` to a reference-architecture SFD, or sweep them in `params()` (for example `'fee_bps': [2.0, 5.0, 10.0]`), to match a venue's costs, make cost a search dimension, or stress-test how sensitive an edge is to the cost assumption. Omitting them keeps the 5 + 5 default, so existing experiments are unchanged.
+
 This makes snapshot backtests fast and comparable across rounds, but it also means they are not trying to be a full execution simulator.
 
 ### Output columns
--- a/limen/sfd/reference_architecture/logreg_binary.py
+++ b/limen/sfd/reference_architecture/logreg_binary.py
@@ -6,6 +6,8 @@ from sklearn.linear_model import LogisticRegression
 
 from limen.calibration import fit_calibrator
 from limen.metrics.binary_metrics import binary_metrics
+from limen.sfd.reference_architecture.base import DEFAULT_FEE_BPS
+from limen.sfd.reference_architecture.base import DEFAULT_SLIP_BPS
 from limen.sfd.reference_architecture.base import ReferenceModel
 
 if TYPE_CHECKING:
@@ -159,8 +161,8 @@ def logreg_binary(data: dict,
                   warm_start: bool = False,
                   n_jobs: int = -1,
                   l1_ratio: float | None = None,
-                  fee_bps: float = 5.0,
-                  slip_bps: float = 5.0,
+                  fee_bps: float = DEFAULT_FEE_BPS,
+                  slip_bps: float = DEFAULT_SLIP_BPS,
                   prediction_calibration_config: 'CalibrationConfig | None' = None) -> dict:
 
     '''
--- a/limen/sfd/reference_architecture/logreg_binary.py
+++ b/limen/sfd/reference_architecture/logreg_binary.py
@@ -159,6 +159,8 @@ def logreg_binary(data: dict,
                   warm_start: bool = False,
                   n_jobs: int = -1,
                   l1_ratio: float | None = None,
+                  fee_bps: float = 5.0,
+                  slip_bps: float = 5.0,
                   prediction_calibration_config: 'CalibrationConfig | None' = None) -> dict:
 
     '''
@@ -182,6 +184,8 @@ def logreg_binary(data: dict,
         warm_start (bool): Whether to reuse previous solution
         n_jobs (int): Number of parallel jobs
         l1_ratio (float | None): Elastic-net mixing parameter
+        fee_bps (float): Per-fill fee in basis points applied in the backtest
+        slip_bps (float): Per-fill slippage in basis points applied in the backtest
         prediction_calibration_config (CalibrationConfig | None): Optional calibration config
 
     Returns:
@@ -208,6 +212,8 @@ def logreg_binary(data: dict,
         l1_ratio=l1_ratio,
     )
 
+    model.fee_bps = fee_bps
+    model.slip_bps = slip_bps
     result = model.evaluate(data, inline_metrics=True)
     result['_model'] = model
     return result
--- a/limen/sfd/reference_architecture/random_binary.py
+++ b/limen/sfd/reference_architecture/random_binary.py
@@ -83,7 +83,9 @@ class RandomBinary(ReferenceModel):
 
 
 def random_binary(data: dict,
-                  random_weights: float = 0.5) -> dict:
+                  random_weights: float = 0.5,
+                  fee_bps: float = 5.0,
+                  slip_bps: float = 5.0) -> dict:
 
     '''
     Random binary classifier for testing and demonstration purposes.
@@ -91,6 +93,8 @@ def random_binary(data: dict,
     Args:
         data (dict): Data dictionary with x_test, y_test
         random_weights (float): Probability weight for class 1 (0.0 to 1.0)
+        fee_bps (float): Per-fill fee in basis points applied in the backtest
+        slip_bps (float): Per-fill slippage in basis points applied in the backtest
 
     Returns:
         dict: Results with binary metrics, predictions, inline confusion metrics,
@@ -98,6 +102,8 @@ def random_binary(data: dict,
     '''
 
     model = RandomBinary().train(data, random_weights=random_weights)
+    model.fee_bps = fee_bps
+    model.slip_bps = slip_bps
     result = model.evaluate(data, inline_metrics=True)
     result['_model'] = model
     return result
--- a/limen/sfd/reference_architecture/random_binary.py
+++ b/limen/sfd/reference_architecture/random_binary.py
@@ -3,6 +3,8 @@ from typing import Any
 import numpy as np
 
 from limen.metrics.binary_metrics import binary_metrics
+from limen.sfd.reference_architecture.base import DEFAULT_FEE_BPS
+from limen.sfd.reference_architecture.base import DEFAULT_SLIP_BPS
 from limen.sfd.reference_architecture.base import ReferenceModel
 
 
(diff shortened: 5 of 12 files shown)


zero-bang (approved):
LGTM. Clean, tightly-scoped slice that does exactly what the PRD asks.

**Verified:**
- All six backtest paths now forward the new knobs: `base._compute_backtest` (used by logreg/random/xgboost/tabpfn) and `RuleBasedStrategy._backtest_split` both pass `fee_bps`/`slip_bps`. No backtest call site was missed.
- `ReferenceModel.__init__` seeds `fee_bps`/`slip_bps` to `DEFAULT_*`, and every subclass either calls `super().__init__()` (logreg/random/tabpfn/rule_based) or inherits it (xgboost has no `__init__`), so the class-based API (`Model().train(...).evaluate(...)`) is safe even when the entry function isn't used.
- Defaults `5.0`/`5.0` match `backtest_snapshot`'s own defaults, so the omit-cost path is genuinely unchanged — `test_omitting_cost_matches_explicit_default` confirms this byte-for-byte (with correct NaN handling).
- The "sweeps from `params()` with no schema change" claim holds: `Manifest.resolve_model_kwargs` (manifest_core.py:721) maps `round_params` onto the function signature by name, so any new signature parameter becomes sweepable automatically. Docs/CHANGELOG claim is accurate.
- Inserting `fee_bps`/`slip_bps` before `prediction_calibration_config` in logreg/tabpfn is safe — no caller passes that arg positionally, and the manifest resolves by name.
- `tests/test_reference_architecture.py` — 22 passed locally.

**Engineering note (non-blocking):** cost is threaded as a post-construction instance mutation (`model.fee_bps = fee_bps`) rather than through `train()`/`__init__` like the other hyperparameters. It works and keeps the diff minimal, but it's a slightly different ownership path than the rest of the model surface — worth keeping in mind if the cost model later grows (e.g. maker/taker asymmetry or per-side fees), at which point folding it into the constructor would read more consistently.

**Forward-looking:** the most predictable next failure mode is sweeping `fee_bps`/`slip_bps` with an architecture exposing a `**kwargs` var-keyword — `resolve_model_kwargs` would route the cost params through the var-keyword branch instead of an explicit signature slot. The current five all have explicit params so this is fine today, but a future architecture relying on `**kwargs` to reach `backtest_snapshot` would need to forward them deliberately.

limen/sfd/reference_architecture/base.py
```diff
@@ -9,6 +9,10 @@
 from limen.log._permutation_confusion_metrics import _confusion_mean_return_pct
 
 
+DEFAULT_FEE_BPS = 5.0
+DEFAULT_SLIP_BPS = 5.0
```
bit-mis, line 14: p3: DEFAULT_FEE_BPS / DEFAULT_SLIP_BPS are introduced here but none of the five entry functions (logreg_binary, random_binary, xgboost_regressor, tabpfn_binary, rule_based) reference them - they all hardcode 5.0 in their signatures (and the omit-vs-explicit test hardcodes 5.0 too). The default therefore lives in ~7 places, so any future change to these constants will silently diverge from the public defaults the constants appear intended to anchor. Consider importing and using them in the signatures, e.g. fee_bps: float = DEFAULT_FEE_BPS, so the default is defined exactly once.

bit-mis (changes requested):

mikkokotila:
Addressed the review feedback in 71e4348:

1. **@bit-mis — single-source default**: the five entry-function signatures now default `fee_bps`/`slip_bps` to `DEFAULT_FEE_BPS` / `DEFAULT_SLIP_BPS` (imported from `base`), and the omit-vs-explicit test uses the constants too. The literal `5.0` now lives in exactly one place.

2. **@Copilot — negative cost**: `ReferenceModel` now exposes `fee_bps`/`slip_bps` as properties with validating setters, so a negative value raises `ValueError` on assignment — in every path (the ML `_compute_backtest` and the `RuleBasedStrategy` backtest), before `backtest_snapshot` is ever called, and regardless of whether price data is present. Added `test_negative_cost_is_rejected`.

Separately (root cause of the earlier CI Tests failure): my first push broke the YAML/foundational architecture-surface tests, which assert the arch signature is covered by the template/foundational param set — `fee_bps`/`slip_bps` are execution params, not model hyperparameters, so I excluded them there (matching how `prediction_calibration_config` is excluded). I also corrected slice #577's Surfaces to list `tests/test_yaml.py`.

`ruff check .` clean; `test_reference_architecture.py` + `test_yaml.py` = 135 passed; full `tests.run` re-run.

bit-mis (approved):

mikkokotila:
On the latest @Copilot round:

- **Non-finite cost** (base.py setters): fixed in 895654a — the `fee_bps`/`slip_bps` setters now reject `nan`/`inf` as well as negatives (`math.isfinite`), and `test_invalid_cost_is_rejected` covers both.
- **Set cost before `train()` to fail fast** (logreg/xgboost/tabpfn): fair point, deferred. An invalid cost is rejected on assignment and fails the round with a clear `ValueError` regardless of ordering; the only difference is one wasted `train()` on an already-misconfigured round. I've kept the assignment after `train()` to keep this diff minimal — happy to reorder as a fast-follow if you'd prefer it.

bit-mis (approved):

mikkokotila:
Superseding this PR. Per discussion, the per-architecture signature-threading is the wrong layer — it couples backtest cost into all five model signatures and leaks it into the model-surface contract (which is exactly what broke the YAML template-surface tests here).

The design is now a **manifest-level backtest config**: `Manifest.set_backtest_config(fee_bps, slip_bps)` + a YAML `backtest:` block, resolved per round (optionally swept via a param reference) and applied at the single backtest call site, with the model architectures untouched. PRD #575 and the slices are updated — #577 (manifest config core) and #579 (YAML block).

Closing; the new design lands via #577 + #579.

Issue #579, opened 2026-06-06
Add a YAML backtest block compiling to the manifest backtest config


mikkokotila:
## Slice ID

S579

## Slice Title

Add a YAML backtest block compiling to the manifest backtest config

## Primary Task Type

operator_surface

## Routing and Dependencies

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: #575
Depends on:
- #577
Unlocks:
- none

## Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `python -c "from limen.experiment.trainer import Trainer; print(Trainer.__name__)"` returns `Trainer`.
> - `python -m tests.run` returns exit code `0`.
> - `ruff check .` returns exit code `0`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `python -c "from limen.yaml import compiler; print(hasattr(compiler, '_apply_backtest'))"` returns `True`
- `python -m pytest tests/test_yaml.py -k test_backtest_block_compiles_to_manifest_config -q` returns exit code `0`
- `python -m pytest tests/test_yaml.py -k test_backtest_block_rejects_invalid_cost -q` returns exit code `0`

## Objective, Baseline, and Scope

> **Significance.** Objective, baseline, and scope define the smallest
> useful change. The objective states the outcome. Baseline is repo or
> runtime state today, expressed as commands and outputs a reader can run
> against `main`. Scope names the exact surfaces this slice may touch.
>
> Accepted:
> - `git show HEAD:pyproject.toml | grep -c 'name = "limen"'` returns `1`.
> - `test -f limen/cli/main.py && echo present` returns `present`.
>
> Rejected:
> - *The system has accumulated technical debt.*
> - *Several paths are confusing.*

[...]

PR #580, opened 2026-06-06
Add a manifest-level backtest cost config (set_backtest_config)

### LINKED ISSUE #577: Add a manifest backtest cost config (set_backtest_config) applied at the backtest call sites

mikkokotila:
Implements slice [#577](https://github.com/Vaquum/Limen/issues/577) (PRD [#575](https://github.com/Vaquum/Limen/issues/575)). Towards #534.

Closes #577

## What

Backtest cost is now configured **on the manifest**, not on each model — the redesign agreed after the per-architecture approach (closed #578) was found to be the wrong layer.

- `Manifest.set_backtest_config(fee_bps, slip_bps)` stores a `BacktestConfig`.
- `run_model` resolves it against the round's params via the existing `_resolve_params` convention (**a number is fixed; a string is a search-param reference**, so `set_backtest_config(fee_bps='fee')` + `params={'fee': [2,5,10]}` sweeps cost), validates it non-negative and finite, and injects `backtest_fee_bps`/`backtest_slip_bps` into the prepared `data` dict — the channel that already carries `price_data_for_backtest`.
- `ReferenceModel._compute_backtest` and `RuleBasedStrategy._backtest_split` forward the cost to `backtest_snapshot` via a shared `_cost_kwargs(data)`, defaulting to `backtest_snapshot`'s own `5.0`/`5.0` when unset.

**The five model architectures are untouched** — `fee_bps`/`slip_bps` never enter the model surface (guarded by a test), so nothing leaks into the template/foundational param-surface contracts.

```python
manifest.set_backtest_config(fee_bps=2.0, slip_bps=3.0)        # fixed
manifest.set_backtest_config(fee_bps='fee', slip_bps='slip')   # swept via params 'fee'/'slip'
```

## Tests (`tests/test_reference_architecture.py`)

- `test_compute_backtest_honours_injected_cost` — injected `backtest_fee_bps=0` → `cost_drag_p50 == 0`; high cost raises drag and lowers net PnL.
- `test_manifest_backtest_config_resolves_and_sweeps` — literal and swept reference resolve and inject; no config → no injection.
- `test_architectures_do_not_expose_cost_parameters` — none of the five entry signatures contain `fee_bps`/`slip_bps` (decoupling guard).
- `test_invalid_backtest_config_is_rejected` — negative / `inf` / `nan` resolved cost raises.

## Scope

Exactly the slice's Surfaces (`manifest_core.py`, `base.py`, `rule_based.py`, the test + `tests/run.py`, `docs/Backtest.md`, CHANGELOG, version → v3.18.0). The four model architectures, `limen/yaml/`, and `backtest_snapshot.py` are untouched (slice Out of Scope). The YAML `backtest:` block follows as slice #579.

## Validation

- `ruff check .` clean; `tests/test_reference_architecture.py` 23 passed; full `tests.run` run locally.

### DIFF
--- a/limen/experiment/manifest_core.py
+++ b/limen/experiment/manifest_core.py
@@ -2,6 +2,8 @@ import copy
 import inspect
 import importlib
 import logging
+import math
+import numbers
 import random
 import re
 from datetime import date
@@ -237,6 +239,15 @@ class DataSourceResolver:
 
 
 @dataclass
+class BacktestConfig:
+
+    '''Backtest cost configuration: per-fill fee and slippage in basis points.'''
+
+    fee_bps: float | str = 5.0
+    slip_bps: float | str = 5.0
+
+
+@dataclass
 class Manifest:
 
     '''Base manifest with shared data pipeline configuration for Loop experiments.'''
@@ -255,6 +266,7 @@ class Manifest:
     architecture_function: Callable = None
     architecture_params: dict[str, ParamValue] = field(default_factory=dict)
     metrics_params: dict[str, ParamValue] = field(default_factory=dict)
+    backtest_config: BacktestConfig | None = None
 
     def _add_transform(self,
                        func: Callable,
@@ -742,6 +754,44 @@ class Manifest:
         return model_kwargs
 
 
+    def set_backtest_config(self,
+                            fee_bps: float | str = 5.0,
+                            slip_bps: float | str = 5.0) -> 'Manifest':
+
+        '''
+        Configure the backtest cost model for this manifest.
+
+        Args:
+            fee_bps (float | str): Per-fill fee in basis points, or a round-param name to sweep
+            slip_bps (float | str): Per-fill slippage in basis points, or a round-param name to sweep
+
+        Returns:
+            Manifest: Self for method chaining
+        '''
+
+        self.backtest_config = BacktestConfig(fee_bps=fee_bps, slip_bps=slip_bps)
+
+        return self
+
+    def _apply_backtest_cost(self, data: dict, round_params: dict[str, Any]) -> None:
+        if self.backtest_config is None:
+            return
+
+        resolved = _resolve_params(
+            {'fee_bps': self.backtest_config.fee_bps, 'slip_bps': self.backtest_config.slip_bps},
+            round_params,
+        )
+        for key in ('fee_bps', 'slip_bps'):
+            value = resolved[key]
+            if (
+                isinstance(value, bool)
+                or not isinstance(value, numbers.Real)
+                or not math.isfinite(value)
+                or value < 0
+            ):
+                raise ValueError(f"Manifest backtest {key} must be a non-negative finite number")
+            data[f"backtest_{key}"] = float(value)
+
     def run_model(self, data: dict, round_params: dict[str, Any]) -> dict:
 
         '''
@@ -764,6 +814,7 @@ class Manifest:
         '''
 
         model_kwargs = self.resolve_model_kwargs(round_params)
+        self._apply_backtest_cost(data, round_params)
         return self.architecture_function(data, **model_kwargs)
 
 
@@ -1030,6 +1081,7 @@ class MLManifest(Manifest):
                     threshold_params=resolved.threshold_params,
                 )
                 model_kwargs['prediction_calibration_config'] = config
+        self._apply_backtest_cost(data, round_params)
         return self.architecture_function(data, **model_kwargs)
 
     def sensor_input_prep(
--- a/limen/experiment/manifest_core.py
+++ b/limen/experiment/manifest_core.py
@@ -777,15 +777,19 @@ class Manifest:
         if self.backtest_config is None:
             return
 
-        resolved = _resolve_params(
-            {'fee_bps': self.backtest_config.fee_bps, 'slip_bps': self.backtest_config.slip_bps},
-            round_params,
-        )
+        raw = {'fee_bps': self.backtest_config.fee_bps, 'slip_bps': self.backtest_config.slip_bps}
+        resolved = _resolve_params(raw, round_params)
         for key in ('fee_bps', 'slip_bps'):
+            original = raw[key]
             value = resolved[key]
-            if isinstance(value, str):
+            if (
+                isinstance(original, str)
+                and isinstance(value, str)
+                and value == original
+                and original not in round_params
+            ):
                 raise ValueError(
-                    f"Manifest backtest {key} references unknown search-param '{value}'; "
+                    f"Manifest backtest {key} references unknown search-param '{original}'; "
                     "add it to params() or pass a number"
                 )
             if (
--- a/limen/experiment/manifest_core.py
+++ b/limen/experiment/manifest_core.py
@@ -794,7 +794,9 @@ class Manifest:
                 or not math.isfinite(value)
                 or value < 0
             ):
-                raise ValueError(f"Manifest backtest {key} must be a non-negative finite number")
+                raise ValueError(
+                    f"Manifest backtest {key} must be a non-negative finite number, got {value!r}"
+                )
             data[f"backtest_{key}"] = float(value)
 
     def run_model(self, data: dict, round_params: dict[str, Any]) -> dict:
--- a/limen/experiment/manifest_core.py
+++ b/limen/experiment/manifest_core.py
@@ -783,6 +783,11 @@ class Manifest:
         )
         for key in ('fee_bps', 'slip_bps'):
             value = resolved[key]
+            if isinstance(value, str):
+                raise ValueError(
+                    f"Manifest backtest {key} references unknown search-param '{value}'; "
+                    "add it to params() or pass a number"
+                )
             if (
                 isinstance(value, bool)
                 or not isinstance(value, numbers.Real)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1064,3 +1064,7 @@ Note: add all new changelog entries to the bottom of this file.
 - Add `vpin` to `limen.features` — Volume-synchronized Probability of Informed Trading (Easley, Lopez de Prado, and O'Hara): the rolling absolute BVC buy/sell imbalance over total volume, a flow-toxicity gauge in `[0, 1]`. Composes `bulk_volume_classification`; feed volume bars for canonical equal-volume buckets
 - Add `cusum_filter` to `limen.features` — Lopez de Prado's symmetric CUSUM event gate on the close log-return path, emitting an Int8 `cusum_event` flag (`1` up, `-1` down, `0` none) that samples only moves whose cumulative magnitude breaches `threshold`
 - Add `time_to_funding` to `limen.features` — continuous hours until the next funding settlement (default cadence every `8` hours from `0` UTC), complementing the existing `is_funding_hour` flag
+
+## v3.18.0 on 6th of June, 2026
+
+- Add a manifest-level backtest cost config: `Manifest.set_backtest_config(fee_bps, slip_bps)` stores a `BacktestConfig` that `run_model` resolves against the round's params (a number is fixed; a string is a search-param reference, so `set_backtest_config(fee_bps='fee')` sweeps cost via `params()`), validates non-negative and finite, and injects into the prepared `data` dict. `ReferenceModel._compute_backtest` and the rule-based backtest forward the resolved cost to `backtest_snapshot`, defaulting to the existing `5.0`/`5.0` when no config is set. The model architectures are untouched — cost stays off the model surface. Towards #534
--- a/docs/Backtest.md
+++ b/docs/Backtest.md
@@ -49,6 +49,8 @@ The current snapshot backtest is intentionally simple and opinionated:
 - output metrics are quantiles over their declared substrate
 - return and ratio outputs are basis-point scaled
 
+The fee and slippage rates default to `5.0` bps each per fill (a ~20 bps round trip) and are configured on the manifest, not on the model: `manifest.set_backtest_config(fee_bps=..., slip_bps=...)`. Each value is a fixed number or a search-param name — pass a param name to sweep cost across the search (for example `set_backtest_config(fee_bps='fee')` with `fee` in `params()`), to match a venue's costs, make cost a search dimension, or stress-test how sensitive an edge is to the cost assumption. Omitting the config keeps the 5 + 5 default, so existing experiments are unchanged.
+
 This makes snapshot backtests fast and comparable across rounds, but it also means they are not trying to be a full execution simulator.
 
 ### Output columns
--- a/limen/sfd/reference_architecture/base.py
+++ b/limen/sfd/reference_architecture/base.py
@@ -122,6 +122,13 @@ class ReferenceModel(ABC):
 
         return results
 
+    def _cost_kwargs(self, data: dict) -> dict:
+        return {
+            key: data[f"backtest_{key}"]
+            for key in ('fee_bps', 'slip_bps')
+            if f"backtest_{key}" in data
+        }
+
     def _compute_backtest(self, preds: np.ndarray, data: dict) -> dict:
 
         '''
@@ -151,7 +158,7 @@ class ReferenceModel(ABC):
 
         bt_input = pd.DataFrame(bt_input_data)
 
-        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1)
+        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1, **self._cost_kwargs(data))
 
         if bt_result.empty:
             return {}
--- a/limen/sfd/reference_architecture/rule_based.py
+++ b/limen/sfd/reference_architecture/rule_based.py
@@ -72,10 +72,11 @@ class RuleBasedStrategy(ReferenceModel):
         strategy = data['strategy']
         cond_index = {c['id']: c for c in strategy['conditions']}
 
+        cost_kwargs = self._cost_kwargs(data)
         for split in ('train', 'val', 'test'):
             pos = self._resolve(cond_index[strategy['entry']], cond_index, data[split]).fill_null(False).to_numpy().astype(int)
             positions[split] = pos
-            backtest_results[split] = self._backtest_split(data[split], pos)
+            backtest_results[split] = self._backtest_split(data[split], pos, cost_kwargs)
 
         results = rule_based_metrics(
             positions,
@@ -108,7 +109,7 @@ class RuleBasedStrategy(ReferenceModel):
             result = result & s if operator == 'and' else result | s
         return result
 
-    def _backtest_split(self, df: pl.DataFrame, positions: np.ndarray) -> dict:
+    def _backtest_split(self, df: pl.DataFrame, positions: np.ndarray, cost_kwargs: dict) -> dict:
         if 'open' not in df.columns or 'close' not in df.columns:
             return {}
         open_arr = df['open'].to_numpy().astype(float)
@@ -123,7 +124,7 @@ class RuleBasedStrategy(ReferenceModel):
             bt_input_data['datetime'] = df['datetime'].to_numpy()
 
         bt_input = pd.DataFrame(bt_input_data)
-        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1)
+        bt_result = backtest_snapshot(bt_input, execution_lag_bars=1, **cost_kwargs)
         if bt_result.empty:
             return {}
         return bt_result.iloc[0].to_dict()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "vaquum_limen"
-version = "3.17.0"
+version = "3.18.0"
 description = "Limen unifies parameter search across machine learning and rule-based strategies, with built-in analytics that show not just what works, but why it works."
 readme = "README.md"
 authors = [
--- a/tests/run.py
+++ b/tests/run.py
@@ -480,6 +480,10 @@ from tests.test_reference_architecture import test_rule_based_or_compound_condit
 from tests.test_reference_architecture import test_rule_based_not_operator
 from tests.test_reference_architecture import test_rule_based_empty_operands_raises
 from tests.test_reference_architecture import test_rule_based_unknown_operator_raises
+from tests.test_reference_architecture import test_compute_backtest_honours_injected_cost
+from tests.test_reference_architecture import test_manifest_backtest_config_resolves_and_sweeps
+from tests.test_reference_architecture import test_architectures_do_not_expose_cost_parameters
+from tests.test_reference_architecture import test_invalid_backtest_config_is_rejected
 from tests.test_manifest_prepare_data import test_price_data_for_backtest_has_ohlc_columns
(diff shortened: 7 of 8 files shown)


limen/experiment/manifest_core.py
```diff
@@ -742,6 +754,44 @@ def resolve_model_kwargs(self, round_params: dict[str, Any]) -> dict[str, Any]:
...
+        )
+        for key in ('fee_bps', 'slip_bps'):
+            value = resolved[key]
+            if (
+                isinstance(value, bool)
+                or not isinstance(value, numbers.Real)
+                or not math.isfinite(value)
+                or value < 0
+            ):
+                raise ValueError(f"Manifest backtest {key} must be a non-negative finite number")
```
zero-bang: Non-blocking DX note. When `fee_bps`/`slip_bps` is a string that doesn't resolve (typo'd param name, or the user forgot to add it to `params()`), `_resolve_params` returns the literal string unchanged, and it lands here as `"must be a non-negative finite number"`. Since string-as-param-reference is the headline capability of this feature, an unresolved reference is the most likely user error, and that message points at the wrong cause. Consider distinguishing the case — e.g. when the resolved value is still a `str`, raise something like `f"Manifest backtest {key} references unknown search-param '{value}'; add it to params() or pass a number"`. Cheap to add and saves a confusing debugging session.

zero-bang (approved):
Approve. Clean, well-scoped slice. The redesign to put cost on the manifest rather than the model surface is the right layer, and the implementation funnels every evaluation path through a single seam: both `Manifest.run_model` and `MLManifest.run_model` call `_apply_backtest_cost`, and `RuleBasedManifest` inherits the base `run_model`, so the rule-based path is covered without an extra override. I checked the other callers (`trainer.py`, `experiment_core.py`, `yaml/profiler.py`) — they all go through `run_model`, so there is no evaluation path that silently bypasses the configured cost and falls back to 5/5. `backtest_snapshot`'s keyword-only `fee_bps`/`slip_bps` defaults match the documented `5.0`/`5.0` fallback, `_cost_kwargs` only forwards keys when present so the unset case correctly inherits those defaults, and the `bool`/`Real`/`isfinite`/`<0` validation is appropriately defensive (including the explicit `bool` rejection). Tests are meaningful — the injected-cost test asserts directional behavior (drag up, net PnL down), not just non-emptiness — and I ran the file locally: 23 passed. One non-blocking DX note inline.

Forward-looking failure mode: the next predictable snag is a user setting `set_backtest_config(fee_bps='fee')` but forgetting to add `fee` to `params()` — see the inline note. Slice #579 (the YAML `backtest:` block) will want to round-trip these string-vs-number semantics through deserialization, where a YAML-quoted number could change a fixed value into an accidental param reference.

mikkokotila:
Good catch @zero-bang — addressed in the latest push. An unresolved `fee_bps`/`slip_bps` string reference (typo'd, or not added to `params()`) now raises `Manifest backtest fee_bps references unknown search-param '<name>'; add it to params() or pass a number` instead of the generic finite-number error, with a test covering it. Valid references and numeric literals are unaffected.

zero-bang:
Confirmed — the unresolved-reference branch raises before the generic finite-number check, with exactly the message I had in mind, and `test_invalid_backtest_config_is_rejected` now asserts the `unknown search-param` text for a typo'd `set_backtest_config(fee_bps='missing_param')`. That's the most likely user error pointing at the right cause now. Thanks for the quick turnaround — my approval stands.

limen/experiment/manifest_core.py
```diff
@@ -742,6 +754,49 @@ def resolve_model_kwargs(self, round_params: dict[str, Any]) -> dict[str, Any]:
...
+        if self.backtest_config is None:
+            return
+
+        resolved = _resolve_params(
+            {'fee_bps': self.backtest_config.fee_bps, 'slip_bps': self.backtest_config.slip_bps},
+            round_params,
+        )
+        for key in ('fee_bps', 'slip_bps'):
+            value = resolved[key]
+            if isinstance(value, str):
```
bit-mis: Validation error message names the class (Manifest) but not the offending value or type. When _resolve_params yields something unexpected (e.g. an int that happens to be a bool, or a string that did not resolve), the user sees "must be a non-negative finite number" with no hint of what was actually passed. Including the bad value would help debugging, e.g. f"Manifest backtest {key} must be a non-negative finite number, got {value!r}".

bit-mis (changes requested):

zero-bang:
Good point @bit-mis — addressed in the latest push. The non-negative/finite validation now appends the offending value to the message:

```
Manifest backtest fee_bps must be a non-negative finite number, got -1.0
```

via `f"... got {value!r}"`, so a bad number, a bool, or any unexpected type now shows what was actually passed. `test_invalid_backtest_config_is_rejected` now also asserts the `got ` fragment is present. (The unresolved-string-reference case still raises its own dedicated `unknown search-param` message before this check.)

mikkokotila:
Addressed @bit-mis — the non-negative-finite error now appends `, got {value!r}`, so a rejected `fee_bps`/`slip_bps` shows the offending value (the unresolved-reference case already names the bad param). The earlier CI Tests red was a transient HuggingFace 429 on the test-data fetch, unrelated to the diff; this push re-runs it.

zero-bang:
Reviewed at `9f0f580` — pulled the branch, ran `tests/test_reference_architecture.py` (23 passed) and `ruff check` on the four changed source/test files (clean).

The two review threads are fully resolved:
- **Unresolved string reference** (zero-bang): the `isinstance(value, str)` branch raises the `references unknown search-param '<name>'; add it to params() or pass a number` message *before* the finite-number check, so the most likely user error (typo'd / un-registered param name) now points at the right cause.
- **Offending value in the error** (bit-mis): the non-negative/finite branch now appends `, got {value!r}`, and `test_invalid_backtest_config_is_rejected` asserts the `got ` fragment, so a bad number/bool/type surfaces what was actually passed.

The design is the right layer: every evaluation path (`Manifest.run_model`, `MLManifest.run_model`, and `RuleBasedManifest` via the inherited base) funnels through `_apply_backtest_cost`, and `_cost_kwargs` only forwards keys when present so the unset case correctly inherits `backtest_snapshot`'s 5/5 defaults. The decoupling guard (`test_architectures_do_not_expose_cost_parameters`) keeps cost off the model surface. Validation order, the explicit `bool` rejection, and the `numbers.Real`/`isfinite` checks are all sound.

One non-blocking note for a follow-up, on the same theme as the two threads above: `_resolve_params` treats a leading-underscore string as an unconditional `round_params[value]` lookup, so `set_backtest_config(fee_bps='_missing')` with `_missing` absent raises a raw `KeyError` rather than the friendly `unknown search-param` message. Unlikely for a cost-param name, and out of scope here — just worth keeping in mind for slice #579 when the YAML `backtest:` block round-trips these string-vs-number semantics.

LGTM. 👍

zero-bang:
Thanks @Copilot — looked at both points.

**Resolved-to-string case (valid, fixed in `d0c2f2c`).** You're right: the previous branch raised `unknown search-param` for *any* resolved string, so `set_backtest_config(fee_bps='fee')` + `params={'fee': 'oops'}` would mislabel a non-numeric param value as an unknown param name. The check now fires the `unknown search-param` message **only** for a genuinely unresolved reference — a string `_resolve_params` returned verbatim because it isn't in `round_params`:

```python
if (isinstance(original, str)
        and isinstance(value, str)
        and value == original
        and original not in round_params):
    raise ValueError(f"Manifest backtest {key} references unknown search-param '{original}'; ...")
```

A param that exists but resolves to a non-numeric value now falls through to the numeric check and surfaces the real cause, e.g. `Manifest backtest fee_bps must be a non-negative finite number, got 'oops'`. `test_invalid_backtest_config_is_rejected` gains a `({'fee_bps': 'fee'}, {'fee': 'oops'})` case asserting the `got ` fragment is present and `unknown search-param` is absent.

**Double quotes (no change).** The continuation `"add it to params() or pass a number"` is implicitly concatenated to a double-quoted f-string, and double-quoted interpolated messages are the prevailing convention in `manifest_core.py` (e.g. the `got {sc!r}` / `got {value!r}` validators and `DataSourceResolver` message). Switching just this string to single quotes would break local consistency, so I've left it.

`tests/test_reference_architecture.py` 23 passed; `ruff check` clean.

limen/experiment/manifest_core.py
mikkokotila: Addressed in 9f0f580 — the numeric-validation error now includes the offending value exactly as suggested:

`f"Manifest backtest {key} must be a non-negative finite number, got {value!r}"`

Follow-up d0c2f2c additionally routes a search-param that *resolves* to a non-numeric value into this same message (named via `got …`) instead of the unknown-reference branch, so the "got <value>" hint now covers both the bad-literal and bad-resolved-value cases. Re-requesting review.

bit-mis (approved):

zero-bang:
Reviewed the full diff at `d0c2f2c` and independently verified: pulled the branch into a clean worktree, ran `tests/test_reference_architecture.py` (23 passed) and `ruff check` on the four changed source/test files (clean).

The design is at the right layer — every evaluation path (`Manifest.run_model`, `MLManifest.run_model`, and `RuleBasedManifest` via the inherited base) funnels through the single `_apply_backtest_cost` seam, and `_cost_kwargs` only forwards keys that are present, so the unset case correctly inherits `backtest_snapshot`'s 5/5 defaults rather than re-asserting them. Keeping `fee_bps`/`slip_bps` off the model surface, guarded by `test_architectures_do_not_expose_cost_parameters`, is the part that makes this safe against the template/foundational param-surface contracts.

On the three resolution branches, the ordering is now correct and each fires for the right cause:
- unresolved reference (`isinstance(original, str) and value == original and original not in round_params`) → `unknown search-param`, the most likely user error;
- a param that exists but resolves to a non-numeric value (e.g. `params={'fee': 'oops'}`) → falls through to the numeric check and surfaces `got 'oops'`, which is the Copilot case fixed in `d0c2f2c` and now pinned by a test;
- bad literal / bool / inf / nan / negative → numeric check with the offending value.

The explicit `bool` rejection ahead of `numbers.Real` and the `math.isfinite` guard are both warranted. Tests assert directional behaviour (drag up, net PnL down) rather than mere non-emptiness, which is the right bar.

One out-of-scope note for slice #579, already on the thread: `_resolve_params` treats a leading-underscore string as an unconditional `round_params[value]` lookup, so `set_backtest_config(fee_bps='_x')` with `_x` absent raises a raw `KeyError` instead of the friendly message — worth folding into the YAML round-trip work, not here.

LGTM. 👍

PR #583, opened 2026-06-06
Add a YAML backtest cost block compiling to set_backtest_config

mikkokotila:
## What

Adds a YAML `backtest:` block that compiles to `Manifest.set_backtest_config`, making the manifest-level backtest cost config (shipped in #580) reachable from CLI/YAML manifests. This is **Slice B** of the cost-model work (#534); Slice A (#580, the Python `set_backtest_config`) is already merged.

The block sits under `sfd.manifest`, sibling to `target:`/`scaler:`:

```yaml
sfd:
  manifest:
    backtest:
      fee_bps: "{fee}"    # swept across the search
      slip_bps: 5.0       # fixed
  params:
    fee: [1.0, 5.0, 10.0]
```

Each cost is a fixed number **or** a `"{param}"` reference into `sfd.params` — `fee_bps: "{fee}"` sweeps cost exactly like indicator/target params (the runtime `_resolve_params` already turns a sole `"{fee}"` into the real float, and `ParamCoverage` already cross-checks the reference).

## Design

- **One shared seam.** Compiled in `_apply_base`, which both `_build_ml_manifest` and `_build_rule_based_manifest` call, so the block works for **`ml` and `rule_based`** with no per-type duplication. The architectures stay untouched — cost remains off the model surface.
- **Optional.** Omitting the block keeps the existing `5.0`/`5.0` default; existing manifests are unchanged.
- **Validated** by a focused `BacktestCostSpec` rule (mirroring `CalibrationCrossRef`): rejects a non-mapping block, an unknown key (hard error — a typo'd `slippage:` must not silently no-op a cost), a negative or non-finite cost, a bare (non-`{param}`) string, and a `"{param}"` reference missing from `sfd.params`.

## Testing

Proven end-to-end with a real CLI manifest:
- `limen validate` ✓ and `limen run --dry-run` ✓
- compiled manifest **sweeps** the fee per round (fee=1/5/10 → `backtest_fee_bps` 1/5/10; slip fixed)
- shared path confirmed for `ml` **and** `rule_based`
- every invalid form rejected with a clear message

9 new tests in `tests/test_yaml.py` (registered in `tests/run.py`); ruff clean; full `tests.run` suite passes locally (exit 0).

Towards #534, completes the YAML surface (#579).

### DIFF
--- a/limen/yaml/compiler.py
+++ b/limen/yaml/compiler.py
@@ -139,6 +139,17 @@ def _apply_base(manifest: Manifest, m: dict[str, Any]) -> None:
             raise ValueError(f"'required_columns' must be a list, got {type(cols).__name__}")
         manifest.set_required_bar_columns(list(cols))
 
+    _apply_backtest(manifest, m)
+
+
+def _apply_backtest(manifest: Manifest, m: dict[str, Any]) -> None:
+
+    backtest = m.get('backtest')
+    if backtest is None:
+        return
+    kwargs = {key: backtest[key] for key in ('fee_bps', 'slip_bps') if key in backtest}
+    manifest.set_backtest_config(**kwargs)
+
 
 def _apply_split(manifest: Manifest, m: dict[str, Any]) -> None:
 
--- a/limen/yaml/compiler.py
+++ b/limen/yaml/compiler.py
@@ -145,7 +145,7 @@ def _apply_base(manifest: Manifest, m: dict[str, Any]) -> None:
 def _apply_backtest(manifest: Manifest, m: dict[str, Any]) -> None:
 
     backtest = m.get('backtest')
-    if backtest is None:
+    if not backtest:
         return
     kwargs = {key: backtest[key] for key in ('fee_bps', 'slip_bps') if key in backtest}
     manifest.set_backtest_config(**kwargs)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1068,3 +1068,7 @@ Note: add all new changelog entries to the bottom of this file.
 ## v3.18.0 on 6th of June, 2026
 
 - Add a manifest-level backtest cost config: `Manifest.set_backtest_config(fee_bps, slip_bps)` stores a `BacktestConfig` that `run_model` resolves against the round's params (a number is fixed; a string is a search-param reference, so `set_backtest_config(fee_bps='fee')` sweeps cost via `params()`), validates non-negative and finite, and injects into the prepared `data` dict. `ReferenceModel._compute_backtest` and the rule-based backtest forward the resolved cost to `backtest_snapshot`, defaulting to the existing `5.0`/`5.0` when no config is set. The model architectures are untouched — cost stays off the model surface. Towards #534
+
+## v3.19.0 on 6th of June, 2026
+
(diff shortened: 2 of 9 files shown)


zero-bang (approved):
LGTM — clean slice that closes the YAML surface for the cost model.

**Verified:**
- Shared `_apply_base` seam reaches both `_build_ml_manifest` and `_build_rule_based_manifest`, so `backtest:` works for `ml` and `rule_based` with no duplication.
- `BacktestCostSpec` rejects every invalid form (non-mapping, unknown key as a hard error, negative/non-finite/bool/list, bare non-`{param}` string, and a `{param}` missing from `sfd.params`); bool-before-numeric ordering is correct.
- `ParamCoverage`'s recursive `_walk_refs` already picks up `"{fee}"` inside the block, so a swept cost param raises no spurious unused-param warning — test confirms.
- Runtime resolution is consistent: the compiler stores the braced `"{fee}"` and `_resolve_params` turns the sole-`{param}` form into the round float.
- Docs example traces correctly against the implementation; CHANGELOG + version bump present; ruff clean; all 121 `test_yaml.py` tests pass (the full `tests.run` failure here is an environmental `libgomp.so.1` miss, unrelated to this diff).

**Engineering note (non-blocking):** the cost-key set `('fee_bps', 'slip_bps')` is now spelled out in three places — `BACKTEST_OPTIONAL` (schema.py:55), the compiler comprehension (compiler.py:150), and the `BacktestCostSpec` loop (rules.py:1023). The next predictable failure mode: a future third cost key (say `min_fee_bps`) added to only one or two of these would silently no-op — schema accepts it but the compiler drops it, or the compiler forwards it but the rule never validates it. Worth a shared constant if a third dimension ever lands.

limen/yaml/compiler.py
```diff
@@ -139,6 +139,17 @@ def _apply_base(manifest: Manifest, m: dict[str, Any]) -> None:
             raise ValueError(f"'required_columns' must be a list, got {type(cols).__name__}")
         manifest.set_required_bar_columns(list(cols))
 
+    _apply_backtest(manifest, m)
+
+
+def _apply_backtest(manifest: Manifest, m: dict[str, Any]) -> None:
+
+    backtest = m.get('backtest')
+    if backtest is None:
```
bit-mis: An empty backtest mapping passes the is-None guard and reaches set_backtest_config with no kwargs, installing a non-None BacktestConfig with defaults instead of the no-block sentinel backtest_config=None that test_build_manifest_no_backtest_block_leaves_config_none asserts. The validator does not catch this since an empty mapping has no keys for BacktestCostSpec or NoUnknownKeys to flag. Consider checking if not backtest so an empty mapping is treated like an omitted block, matching the documented contract that omitting keeps the 5/5 default.

bit-mis (changes requested):

limen/yaml/compiler.py
mikkokotila: Fixed in 9a18a4c exactly as suggested — the guard is now `if not backtest:`, so an empty `backtest: {}` mapping is treated like an omitted block and leaves `backtest_config=None` (matching the documented 5/5-default contract). Added `test_build_manifest_empty_backtest_block_leaves_config_none` and documented the empty-block behaviour in `docs/Backtest.md`. Re-requesting review.

bit-mis (approved):

Issue #582, opened 2026-06-06
Replace the calendar-bound backtest snapshot ledger with a purely bar-based three-set report


mikkokotila:
## Slice ID

S582

## Slice Title

Replace the calendar-bound backtest snapshot ledger with a purely bar-based report

## Primary Task Type

correctness

## Routing and Dependencies

> **Significance.** Routing is the minimum information needed to place this
> slice in the dependency graph. Parent and dependencies must be existing
> issue numbers; unlocks may be forward-looking. Prose description of the
> relationship is not accepted — only issue references. This section is not
> a planning document; it is a graph edge list.

Parent issue: #581
Depends on:
- #534
Unlocks:
- Vaquum/Furnace#76

## Capability

> **Significance.** A slice delivers exactly one capability. The capability
> must be expressible as a set of assertions that become mechanically true
> at merge. This is the answer to *what changes in the system that an
> external observer can verify by running a command?* Every line below
> must be evaluable as a shell command, SQL query, Python expression, or
> HTTP request that returns a definite result. One assertion per line.
>
> Accepted:
> - `python -c "from limen.experiment.trainer import Trainer; print(Trainer.__name__)"` returns `Trainer`.
> - `python -m tests.run` returns exit code `0`.
> - `ruff check .` returns exit code `0`.
>
> Rejected:
> - *Freeze the current-truth read contract.*
> - *Strict CQRS separation.*
> - *Terminal proof obligations are complete.*

Assertions (one per line, each executable as-is):
- `python -c "from limen.backtest.backtest_snapshot import BACKTEST_SNAPSHOT_COLUMNS as c; print(len(c))"` returns `21`
- `python -c "from limen.backtest.backtest_snapshot import BACKTEST_SNAPSHOT_COLUMNS as c; print(c == ['edge_bps_p5','edge_bps_p50','edge_bps_p95','pnl_bps_p5','pnl_bps_p50','pnl_bps_p95','cost_bps_p5','cost_bps_p50','cost_bps_p95','drawdown_bps_p5','drawdown_bps_p50','drawdown_bps_p95','win_rate','pnl_per_bar_bps','avg_win_bps','avg_loss_bps','cvar_95_pnl_bps','trades_per_bar','in_market_per_bar','inventory_per_bar','cost_per_bar_bps'])"` returns `True` (exact set and order — subsumes presence and no-legacy-names)
[...]

Issue #581, opened 2026-06-06
Make the backtest snapshot report purely bar-based (no wall-clock unit anywhere)


mikkokotila:
## Objective

The backtest snapshot's reported metric ledger silently binds three of its eight metric families to wall-clock calendar days (`clock_window='1D'`, `dt.floor('1D')`) and a fourth to elapsed wall-clock days. Limen runs on information-driven and sub-daily bars (dollar, volume, imbalance, 15m). On those bars, calendar bucketing is incoherent and **fails silently**: bars-per-day vary with activity, a single bar can span days, and the grouping re-introduces the heteroscedasticity activity bars exist to remove. The report must become **purely bar-based — one unit (the bar), one population (all bars in the window), no concept of time, and every column run-length-invariant (intensive).**

## Summary

Rewrite `backtest_snapshot` so every column is a per-bar quantity over all bars in the window — no `datetime`, no calendar, no per-trade/fill/entry subsetting. Four per-bar **distributions** (`p5/p50/p95`) and nine intensive **scalars**. Drop the calendar metrics (`rolling_return_net_bps`, `return_on_exposure`, `drawdown_duration_days`) and recompute the survivors in bar space; add per-bar return/cost rates, exposure, turnover, and the win/loss decomposition. Update every Limen consumer (cohort selectors, rule-based metrics, tests, docs) and the downstream Furnace UI in lock-step. Single PR.

## Is This Really Necessary?

Yes, now. The calendar binding produces wrong, non-comparable numbers on exactly the bar types Limen is built for, silently. Those columns feed cohort selection (`backtest_pareto`, `diverse_metrics`) and the Furnace results UI, so the defect propagates into ranking and display. The snapshot is already being reshaped by the cost-params work (#534); folding the bar-native rewrite, the rename, and the new metrics into one cut breaks the downstream contract once, not repeatedly.

## Named Concepts

- **Bar** — the only unit. **Population** — all bars in the window (inaction included: a flat bar is a real `0`, not excluded). No wall-clock time.
- **Intensive** — every column is a rate, ratio, or per-bar quantity; nothing scales with run length (which is arbitrary, a week to a year). No counts, no cumulative totals.
- **Distribution vs scalar** — four metrics carry `p5/p50/p95`; the rest are single intensive scalars.
[...]

PR #584, opened 2026-06-07
Replace calendar-bound backtest snapshot with a bar-based ledger

mikkokotila:
## What

Replaces the backtest `snapshot` metric ledger — previously partly bound to wall-clock calendar days — with a **purely bar-based** one: one unit (the bar), one population (every bar in the window, with a flat bar counted as a real `0`), and every column intensive (run-length-invariant). No `datetime`/calendar anywhere.

Implements slice #582 (PRD #581).

## Ledger (21 columns)

**Per-bar distributions** (`p5`/`p50`/`p95`): `edge_bps` (gross), `pnl_bps` (net), `cost_bps` (gross − net), `drawdown_bps` (net equity vs its running peak).

**Intensive scalars**: `win_rate`, `pnl_per_bar_bps`, `avg_win_bps`, `avg_loss_bps`, `cvar_95_pnl_bps`, `trades_per_bar`, `in_market_per_bar`, `inventory_per_bar`, `cost_per_bar_bps`.

Removed: `rolling_return_net_bps`, `return_on_exposure`, `drawdown_duration_days` (calendar / wall-clock) and the `datetime_col` / `clock_window` parameters.

## Notes

- **Atoms unchanged**: `pos`, `R_gross`, `R_net`, `eq_net` compute identically; the time machinery and dead per-trade code (`run_ids`, `trade_pnl_*`) are removed.
- **Sentinels**: `avg_win_bps`/`avg_loss_bps` are `NaN` when there are no winning/losing bars; `cvar_95_pnl_bps` is `NaN` below 20 bars, else the mean of the lowest `floor(0.05·N)` per-bar net returns (sorted prefix, so zero-ties can't inflate past 5%).
- **Cohort selectors** swap `return_on_exposure_p50 → pnl_per_bar_bps` — a deliberate ranking change (exposure-normalized → per-bar return), since `roe` is gone.
- **`drawdown_bps`** is the one distribution whose deep tail drifts with horizon (max-drawdown grows with sample length) — documented and deliberate; the run-length-invariance test asserts only the 8 truly-intensive scalars.
- **Version `3.20.0`** (3.19.0 was already taken by the merged cost-params work).

## Downstream

Breaking change to the column contract — **Vaquum/Furnace#76** adapts its UI in lock-step and bumps its `vaquum_limen` pin to `>= 3.20.0`.

## Validation

- `ruff check .` clean; all affected and new tests pass locally — the column contract, sentinels, CVaR tail rule, window-doubling invariance, time-free check, and a repo-wide no-legacy-names scan.
- Reviewed by two independent CLIs (Codex + Claude); their findings folded in (CVaR `N = total_bars`; cost test moved to the `cost_per_bar_bps`/`pnl_per_bar_bps` means; the `drawdown_std_bps` rule-based assertion updated for the bar-based semantics).
- The only known failing test is `test_tabpfn`, which is pre-existing on `main` and unrelated (tabpfn is not touched here).


docs/Cohort.md
```diff
@@ -135,7 +135,7 @@ Use them by built-in name:
 Cohort(
     experiment_log_path='experiments/my_exp',
     selector='top_n',
-    selector_params={'column': 'backtest_trade_pnl_net_bps_p50', 'n': 5},
+    selector_params={'column': 'backtest_pnl_bps_p50', 'n': 5},
```
zero-bang: This `top_n` example now produces a degenerate ranking for the common case. `backtest_pnl_bps_p50` is the median net return over **all** bars, and it is exactly `0.0` for any strategy in-market less than 50% of bars (verified: an ~18.5%-in-market run gives `pnl_bps_p50 == 0.0`). With every cohort row tied at `0.0`, `top_n` falls through to the `_id_sort_key` tie-break and effectively selects by lowest id — not by performance. The old `backtest_trade_pnl_net_bps_p50` was a per-trade median and stayed informative. Recommend pointing the example (here and line 152) at an intensive scalar such as `backtest_pnl_per_bar_bps`, which carries the real per-bar signal.

limen/cohort/sfc/backtest_pareto.py
```diff
@@ -17,11 +17,11 @@ def select(context: dict[str, Any],
         raise ValueError('backtest_pareto min_signals must be a non-negative integer')
 
     default_metrics = [
-        'backtest_trade_pnl_net_bps_p50',
-        'backtest_edge_per_signal_bps_p50',
-        'backtest_return_on_exposure_p50',
-        'backtest_drawdown_depth_bps_p50',
-        'backtest_cvar_95_return_bps',
+        'backtest_pnl_bps_p50',
```
zero-bang: Two of these five Pareto dimensions are inert for the typical sparse strategy: `backtest_pnl_bps_p50` and `backtest_edge_bps_p50` are medians over all bars and collapse to `0.0` whenever in-market < 50% of bars (the norm here). A constant column neither helps nor blocks domination, and in the score-normalization step `hi == lo` makes it contribute a flat `1.0` to every row — so the front effectively ranks on only the 3 remaining metrics while advertising 5. Consider swapping the two `_p50` distribution medians for the intensive scalars (e.g. `backtest_pnl_per_bar_bps` is already here; `backtest_avg_win_bps`/`backtest_win_rate`) or for the tail quantiles (`_p95`/`_p5`), which stay discriminating under sparse positions. Same applies to the default group in `diverse_metrics.py`.

zero-bang (changes requested):
Solid, well-tested rewrite: the atoms (`pos`/`R_gross`/`R_net`/`eq_net`) are untouched, the time machinery and dead per-trade code are cleanly excised, the new sentinels/CVaR rule are sensible, and the test suite is genuinely strong (population, sentinels, CVaR tail, window-doubling invariance, time-free scan, repo-wide legacy-name scan). I ran the affected modules locally and they pass. The legacy-name scan confirms no stale `edge_per_signal`/`trade_pnl_net`/etc. references remain in `limen/` or `docs/` (the surviving `datetime_col` hits are unrelated feature modules).

One substantive concern, not in the lines that changed but in how the new columns are *consumed* for ranking: the per-bar distribution **medians** (`*_p50`) are structurally `0.0` for any strategy in-market less than 50% of bars — which is the norm for these long-only signal strategies. I confirmed empirically (an ~18.5%-in-market run yields `edge_bps_p50 = pnl_bps_p50 = cost_bps_p50 = 0.0`). The distribution columns are fine for display, but using `_p50` as a *selector key* silently neutralizes those dimensions. Details inline. Marking request-changes for the `top_n` doc example, which becomes a degenerate all-zero ranking.

Engineering note (no action required): `inventory_per_bar` is `pos.astype(float).mean()` and `in_market_per_bar` is `(pos != 0).mean()` — under the long-only all-in model these are identical by construction, so the ledger carries two columns that can never differ until fractional sizing exists. Forward-looking failure mode to watch: the `drawdown_bps` deep tail (`p5`) drifts upward with window length (acknowledged in the PR), and `rule_based_metrics` feeds `drawdown_bps_p5` into `drawdown_std_bps` across train/val/test — once splits differ materially in length, that std will reflect horizon, not instability, when the replacement stability rule is eventually defined against it.

docs/Cohort.md
mikkokotila: Fixed in 1d5a5b6 — both `top_n` examples (here and line 152) now point at `backtest_pnl_per_bar_bps`, the intensive per-bar scalar, instead of the all-bars median.

limen/cohort/sfc/backtest_pareto.py
mikkokotila: Fixed in 1d5a5b6. Both selectors now key on the discriminating set — `pnl_per_bar_bps`, `edge_bps_p95`, `win_rate`, `drawdown_bps_p5`, `cvar_95_pnl_bps` — so no dimension collapses to a constant under sparse positions. I chose `win_rate` over `avg_win_bps`: `avg_win_bps` is `NaN` on a no-win round and both selectors call `dropna(subset=metric_cols)`, which would silently drop those candidates; `win_rate` is a fraction over all bars and is `0.0` (never `NaN`) instead. Applied identically to the default group in `diverse_metrics.py`.

tests/test_metrics_and_log_helpers.py
```diff
@@ -400,9 +406,148 @@ def test_backtest_snapshot_drops_predictions_without_immediate_next_execution_ba
...
+
+def test_no_legacy_backtest_column_names() -> None:
+    repo_root = Path(__file__).resolve().parent.parent
+    legacy = re.compile(
+        r'edge_per_signal|trade_pnl_net|cost_drag|rolling_return|return_on_exposure'
+        r'|drawdown_depth|drawdown_duration|cvar_95_return|bar_in_market|capital_in_market'
+    )
+    offenders = [
+        str(path.relative_to(repo_root))
+        for base in ('limen', 'docs')
```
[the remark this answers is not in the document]
mikkokotila: Fixed in 1d5a5b6 — the scan bases are now `('limen', 'docs', 'tests')`, with only the test file itself excluded via `path.resolve() != this_file` (it contains the legacy regex literally, so it would otherwise self-trip).

mikkokotila:
Thanks for the careful pass. Addressed in 1d5a5b6:

- **`*_p50` as selector keys (request-changes):** both `backtest_pareto` and `diverse_metrics` now key on the discriminating set `[pnl_per_bar_bps, edge_bps_p95, win_rate, drawdown_bps_p5, cvar_95_pnl_bps]` — no all-bars medians, so nothing collapses to a constant under sub-50%-in-market positions. The `top_n` doc examples point at `pnl_per_bar_bps`.
- **`win_rate` over `avg_win_bps`:** `avg_win_bps` is `NaN` on no-win rounds and the selectors `dropna(subset=metric_cols)`, which would silently drop those candidates; `win_rate` never goes `NaN`.
- **Legacy-name scan (Copilot):** now scans `tests/` too, excluding only the test file itself.

CHANGELOG clause updated to record the discriminating-set choice. Full suite green locally (933 passed); both review CLIs confirm the relabels preserve the two selector-test outcomes (min-max / StandardScaler normalization is invariant to the monotonic value rescale).

On the two no-action engineering notes:

- **`inventory_per_bar` vs `in_market_per_bar`:** correct that they coincide under the long-only all-in model (`capital_fraction` is `{0,1}`). They are kept as distinct columns deliberately — occupancy (binary) vs. deployed capital (fraction) — and diverge the moment fractional sizing exists. Holding both now keeps the contract stable for that.
- **`drawdown_bps_p5` horizon drift into `drawdown_std_bps`:** noted, and left as-is for this PR. The replacement stability rule isn't defined here; when it is, that's the right place to make it length-invariant rather than feeding a horizon-sensitive tail into a cross-split std.

zero-bang (approved):

mikkokotila:
Follow-up on the two engineering notes (post-approval, non-blocking — flagging for the record):

**Note 2 — `in_market_per_bar` ≡ `inventory_per_bar`:** confirmed byte-identical (both reduce to `pos.sum()/total_bars`, since `capital_fraction = pos.astype(float)`), and they cannot diverge under the all-in model. Collapsed to one column — kept `inventory_per_bar` (average position held per bar, the name that generalizes to a deployed fraction once sizing exists) and dropped `in_market_per_bar`. Ledger 21 → 20. The removed name is locked into the legacy-name guard so it can't reappear. (97b35de)

**Note 3 — `win_rate` semantics:** documented in the docstring and `Backtest.md` — it is the share of *all* bars with positive net return (a flat bar is not a win), so it cannot exceed the in-market share, which equals `inventory_per_bar` **under the all-in model** (the bound is model-scoped: under fractional sizing `win_rate` can exceed `inventory_per_bar`).

Furnace updated to the 20-column contract (Vaquum/Furnace#76). Both review CLIs ran on the diff and caught (a) an unscoped `win_rate` bound and (b) a stale "21" in two doc headers — both fixed before push. Full suite green locally.

mikkokotila:
Renamed the snapshot ledger column `win_rate` → **`wins_per_bar`** (prefixed form `backtest_wins_per_bar`). It computes winning bars ÷ all bars, not the conventional wins ÷ trades, so the old name implied the wrong denominator — and `backtest_sequential` already exposes a genuine trade `win_rate`, so the two collided by name. `wins_per_bar` states the denominator and matches the per-bar family (`pnl_per_bar_bps`, `trades_per_bar`, `inventory_per_bar`, `cost_per_bar_bps`).

`backtest_sequential`'s `win_rate` (wins ÷ total_trades) is unrelated and left untouched. Furnace contract synced (Vaquum/Furnace#76). Both review CLIs returned SAFE; full suite green locally (933 passed). (2d52b83)

PR #585, opened 2026-06-08
Remove the stateful BacktestSequential backtest entirely

mikkokotila:
## What

Removes the stateful `BacktestSequential` backtest from the package, as a follow-up to the bar-based snapshot work (#584). The vectorized `backtest_snapshot` is now Limen's only backtest surface.

## Changes

- **Delete** `limen/backtest/backtest_sequential.py` and its `tests/test_backtest_conviction.py` suite.
- **Drop the public export** `limen.BacktestSequential` — both the import and the `__all__` entry in `limen/__init__.py`, plus its registration in `tests/run.py`.
- **Docs:** prune the "Sequential Backtest" / "Sequential Ledger Semantics" sections from `docs/Backtest.md` (and de-dualise the intro), and the sequential references in `limen/backtest/README.md` and `limen/trading/README.md`.
- Version `3.20.0` → `3.21.0`; CHANGELOG entry added.

## Out of scope / preserved

- **`Account` / `limen.trading` stays.** `Account` is a public export (`limen.Account`) with its own `test_account_conviction` suite; `BacktestSequential` was just one consumer, so `Account` is not orphaned in a way that forces deletion. The `trading` README now describes `Account` as a standalone primitive with no in-package consumer. *(If you'd rather also retire `Account`, that's a separate call — say the word and I'll do it in a follow-up.)*
- The unrelated `split_sequential` data-splitter and generic "sequential" wording are untouched.
- Historical CHANGELOG entries mentioning `BacktestSequential` are left intact.

## Breaking change

`limen.BacktestSequential` is removed. Callers must migrate to `backtest_snapshot`. An org-wide code search found **no downstream consumer** (Furnace or otherwise), so no lock-step coordination is required.

## Verification

- `ruff check .` clean; `import limen` works (`BacktestSequential` gone, `Account` present).
- Full suite: **932 passed, 0 failed** (down exactly 1 — the single registered conviction test).
- Reviewed with both Codex and Claude CLIs — both **SAFE**, no dangling references across the tree.


zero-bang (approved):
LGTM — clean, internally-consistent deletion. Verified on the branch:

- `git grep` (excluding CHANGELOG) finds **zero** remaining `BacktestSequential` / `backtest_sequential` / `test_backtest_conviction` references across source, tests, and docs.
- `import limen` works; `BacktestSequential` gone from `dir(limen)`, `Account` still present.
- Confirmed no in-package module consumes `Account` anymore, so the updated `limen/trading/README.md` wording is accurate.
- Repo conventions honored: version `3.20.0 → 3.21.0`, CHANGELOG appended at the bottom, historical `BacktestSequential` mentions left intact, unrelated `split_sequential` untouched.

**Engineering note:** keeping `Account` is the right call — it's a public export with its own conviction suite, so this PR isn't forced to also retire it. But be aware it's now an orphaned primitive (no in-package consumer; only its own test keeps it exercised). It will predictably resurface in any future dead-code audit, so the explicit "say the word and I'll retire it" offer in the description is the right framing to resolve sooner rather than later.

**Forward-looking:** the one claim I can't verify from here is "org-wide search found no downstream consumer." The next predictable failure mode is a consumer *outside* the searched scope (e.g. a notebook, a script, or a repo not in the org sweep) importing `limen.BacktestSequential` and hitting an `ImportError` after the version bump — worth a quick sanity ping to anyone who's used the sequential path before this lands.

PR #586, opened 2026-06-08
Remove the Account ledger and limen.trading package

mikkokotila:
## What

Follow-up to #585 (which removed `BacktestSequential` in v3.21.0). `BacktestSequential` was the **only** consumer of the `Account` ledger, so with it gone `Account` / the whole `limen.trading` package is orphaned. This removes it.

## Why now

- **No in-package consumer:** the one remaining backtest, `backtest_snapshot`, is vectorized and computes returns directly — it never uses a stateful ledger.
- **No downstream consumer:** an org-wide code search for `limen.Account` / `from limen.trading` returns nothing outside this repo.
- `Account` was exported + self-tested, but that's circular once nothing uses it. If a stateful backtest ever returns, a purpose-built ledger can return with it.

## Changes

- **Delete** the `limen/trading/` package (`account.py`, `__init__.py`, `README.md`) and `tests/test_account_conviction.py`.
- **Drop the public export** `limen.Account` (import + `__all__` in `limen/__init__.py`) and its `tests/run.py` registration.
- Version `3.21.0` → `3.22.0`; new CHANGELOG entry (the shipped v3.21.0 entry is left intact).

## Breaking change

`limen.Account` is removed. No downstream consumer found, so no lock-step coordination required.

## Verification

- `ruff check .` clean; `import limen` works (`Account` gone); repo-wide scan shows no dangling `Account` / `limen.trading` references (historical CHANGELOG mentions preserved).
- Full suite: **931 passed, 0 failed** (down exactly 1 — the registered `test_account_conviction`).
- Reviewed with the Codex and Claude CLIs across the deletion and the final diff — **SAFE**.


limen/trading/README.md
```diff
@@ -1,43 +0,0 @@
-# `limen.trading`
```
zero-bang, line 1: Deleting this README breaks the docs-site build. `docs-site/scripts/assemble-docs.mjs` still lists it as a source:

```js
{
  source: 'limen/trading/README.md',
  dest: 'packages/trading.md',
  slug: '/packages/trading',
  sidebarPosition: 12,
  sidebarLabel: 'Trading',
},
```

`copyDoc()` reads each source with an unguarded `await fs.readFile(sourcePath, 'utf8')` and `main()` loops over every entry, so a missing file throws `ENOENT` and aborts assembly. Remove this object from the `docs[]` array in `assemble-docs.mjs` (lines 300-306) as part of this PR, and renumber `sidebarPosition` for the following `utils` entry (12 instead of 13) to keep the sequence contiguous.

zero-bang (changes requested):
Thorough, well-scoped removal — the package, its test, the public export, the `tests/run.py` registration, the version bump, and the CHANGELOG entry are all consistent, and a repo-wide scan confirms no other in-package or doc consumer of `Account` / `limen.trading` survives. One blocking miss: a downstream consumer outside the Python package was not updated.

**Blocking:** Deleting `limen/trading/README.md` leaves a dangling source entry in `docs-site/scripts/assemble-docs.mjs` (the `docs[]` array, line 301-306, slug `/packages/trading`). `copyDoc()` does an unguarded `fs.readFile(sourcePath, 'utf8')` (line 515) with no try/catch, and `main()` iterates every doc — so once the README is gone the docs-site assembly throws `ENOENT` and aborts the build. The grep for `limen.trading` that the PR description ran against the repo missed this because it lives in a `.mjs` file referencing the path as a string. See the inline comment for the exact fix.

Once that entry is removed the change is clean. Forward-looking: the next predictable failure mode is the trading entry's `sidebarPosition: 12` leaving a gap before utils (`13`) — cosmetic only for Docusaurus, but worth renumbering while you're in that file.

limen/trading/README.md
mikkokotila: Fixed in be7f0c8 — removed the `limen/trading/README.md` object from the `docs[]` array in `assemble-docs.mjs` and renumbered `utils` 13 → 12 (backtest stays 11), so the sidebar sequence is contiguous. Verified: `node --check` passes and every remaining `source:` path resolves on disk (no ENOENT). Good catch — my dangling-ref scan only covered `.py/.md/.toml/.yaml`, so a path referenced as a string in a `.mjs` slipped through; I re-ran it across all file types and this was the only remaining reference.

CHANGELOG.md
```diff
@@ -1080,3 +1080,7 @@ Note: add all new changelog entries to the bottom of this file.
 ## v3.21.0 on 8th of June, 2026
 
 - Remove the stateful `BacktestSequential` backtest entirely — delete `limen/backtest/backtest_sequential.py` and its `test_backtest_conviction` suite, drop the `limen.BacktestSequential` public export, and prune the sequential sections from `docs/Backtest.md` and the `limen.backtest`/`limen.trading` READMEs. The vectorized `backtest_snapshot` is now Limen's only backtest surface. The `Account` ledger primitive (`limen.Account`) is unchanged — still exported and still covered by `test_account_conviction`. Breaking: callers of `limen.BacktestSequential` must migrate to `backtest_snapshot`.
+
+## v3.22.0 on 8th of June, 2026
+
+- Remove the `Account` ledger and the `limen.trading` package. `BacktestSequential` (removed in 3.21.0) was its only consumer; with the sequential simulator gone, `Account` has no in-package or downstream consumer, and the vectorized `backtest_snapshot` computes returns directly without a ledger. Deletes `limen/trading/` (`account.py`, its `__init__`, and README) and `tests/test_account_conviction`, and drops the `limen.Account` public export. Breaking: callers of `limen.Account` must migrate to `backtest_snapshot`.
```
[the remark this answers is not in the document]
mikkokotila: Fixed in be7f0c8 — the CHANGELOG no longer says to migrate to `backtest_snapshot`. It now states `limen.Account` is removed with no replacement, and explicitly notes `backtest_snapshot` is a bar-based metrics aggregator, not a ledger.

mikkokotila:
Both addressed in be7f0c8:

- **docs-site build (blocking):** removed the `limen.trading` entry from `assemble-docs.mjs` and renumbered the following `utils` entry 13 → 12 (sequence now contiguous: …backtest 11, utils 12). `node --check` passes and every remaining `source:` path resolves on disk. Re your forward-looking note — the sidebarPosition gap is closed by that renumber.
- **CHANGELOG wording:** dropped the misleading `backtest_snapshot` migration claim; it now states `limen.Account` is removed with no replacement (and that `backtest_snapshot` is a metrics aggregator, not a ledger).

On versioning: keeping this a **3.22.0 minor** bump, not a major. This repo ships breaking changes as minor bumps — both the 3.20.0 ledger rewrite and the 3.21.0 sequential-backtest removal were breaking and went out as minor — and an org-wide search confirms no downstream consumer of `limen.Account`, so there's no real-world break to gate a major release on.

zero-bang (approved):

PR #587, opened 2026-06-08
Remove two orphaned feature modules

mikkokotila:
## What

Removes two dead feature modules found while auditing the package for orphan code: `limen/features/hours_since_big_move.py` and `limen/features/hours_since_quantile_line.py`.

## Why they're orphans

- **Not registered** in `limen/features/__init__.py` (the feature library namespace).
- **Referenced by nothing live** — no other feature module, no test, no YAML template, no doc, no docs-site entry. The only mention anywhere is one historical CHANGELOG line.
- **Unreachable via the standard dispatch:** a manifest references a feature as `func: limen.features.<name>`, which the resolver (`limen/yaml/resolver.py`) resolves from the `limen.features` package namespace — an unregistered module isn't an attribute there, so it doesn't resolve.
- **No downstream use:** org-wide code search found no reference outside this repo.

They've simply been carried along passively through past refactors.

## Audit method (how I found them, and what I deliberately kept)

I scanned every module in `limen/` for any reference (import / dotted-path dispatch / re-export / test / template / doc), excluding self and historical CHANGELOG mentions. Exactly two modules came back with zero references. 28 other feature files are also absent from `__init__` but are **kept** — they're genuinely used via composition (e.g. `entry_score_microstructure`, `dynamic_target`), tests (`test_feature_library_*`), or templates (`sma_ratios`, `volume_ratio`), so they are not orphans.

## Verification

- `ruff check .` clean; `import limen` / `import limen.features` work; no residual references in `limen/`, `tests/`, `docs/`, `docs-site/`.
- Full suite: **931 passed, 0 failed** (no test referenced these files, so the count is unchanged from `main`).
- Reviewed with the Codex and Claude CLIs, both verifying orphan status against the live tree — both **SAFE**. (Both noted a non-standard 4-part dotted path could technically import an unregistered module; the CHANGELOG wording reflects this precisely, and no such reference exists anywhere.)

Version `3.22.0` → `3.23.0`.


zero-bang (approved):
LGTM — verified independently against the live tree.

- Confirmed neither module is registered: `limen/features/__init__.py` uses explicit per-module imports + explicit `__all__` (no wildcard/pkgutil discovery), and `hasattr(limen.features, ...)` is `False` for both.
- Repo-wide sweep (code, yaml/json/toml, md/rst) finds only the two files themselves and one historical CHANGELOG line — nothing in tests/, docs/, docs-site/, or any manifest.
- Traced `yaml/resolver.py`: the standard 3-part `func: limen.features.<name>` dispatch does `import limen.features` then `getattr`, which raises AttributeError for an unregistered submodule → unresolvable. Only an explicit 4-part path could import them directly, and none exists anywhere. The CHANGELOG wording reflects this precisely.
- Version bump (3.22.0 → 3.23.0) and CHANGELOG ordering (appended at bottom) are correct.

Clean, accurate, well-scoped deletion with no consumers to break.

Non-blocking forward-looking note: the latent risk isn't these files but the resolver's 4-part fallback — it keeps *any* .py under an allowed namespace importable regardless of `__init__` registration, so `__init__.py` is a convention, not an enforced boundary. A future auto-discovery tool or hand-written 4-part manifest path could silently resurrect an orphan. If you want `__init__` to be the source of truth, a test asserting every `limen/features/*.py` function is exported in `__all__` would turn that convention into a gate.

mikkokotila:
Closing — decision is to **keep** these two features and integrate them properly (register in `limen.features.__init__`, add tests + docs) rather than remove them. They're real, maintained, sensibly-implemented quant features; "unreferenced in-repo" was a weak deletion criterion for a composable feature library. Integration to follow in a separate change.

PR #588, opened 2026-06-08
Bring the line-based context feature family into the library

mikkokotila:
## What

A deliberate consistency pass that makes the **line-based context feature family** first-class members of the library, instead of leaving them implemented-but-unregistered (the earlier #587 deletion is closed — we keep them):

- `active_lines`, `active_quantile_count`, `quantile_line_density`, `hours_since_big_move`, `hours_since_quantile_line`

## Changes

- **Register all five** in `limen/features/__init__.py` (imports + `__all__`), so `from limen.features import <name>` works like every other feature.
- **Document them** in `docs/Features.md` under a new "Line-Based Context Features" section.
- **Add the missing tests** for the two previously-untested members (`hours_since_*`) in `test_feature_library_context.py` (reset-at-line-end, lookback-cap, no-lines/empty-frame), matching the three siblings' pattern; registered in `tests/run.py`.
- Version `3.22.0` → `3.23.0` + CHANGELOG.

## Design note (and one honest tradeoff)

These features take **pre-computed line structures** (`long_lines`/`short_lines` as `list[dict]` with `start_idx`/`end_idx`), not YAML scalar params — so they're composed programmatically, not dispatched via a manifest `func:`. The docs section says this explicitly.

**Tradeoff surfaced in review:** registering them in `__init__` also makes `func: limen.features.<name>` *resolve* through the YAML resolver. The validator only checks that `func` resolves (not that required positional args are satisfiable), so a user who ignores the docs note and references one of these in a manifest would now pass validation and hit a runtime `TypeError` (missing `long_lines`), rather than the previous clean "not resolvable" error. Severity is low (debuggable error, no silent corruption; no template or internal code does this), and the docs note is the guardrail. If you want it *enforced* rather than documented, a small follow-up could add a validator rule that rejects these from `func:` — out of scope here.

## Verification

- Full suite from a clean checkout: **935 passed, 0 failed**.
- `ruff check .` clean; all five import and appear in `limen.features.__all__`; no feature-module logic changed (only registration).
- Reviewed with Codex and Claude CLIs — both verified the registration/docs/tests are correct; they split only on the severity of the resolvability tradeoff above (Claude: low/non-blocking; Codex: medium/flag).

> Developed in an isolated worktree; no in-progress local work touched. `3.23.0` is the next free version off `main`; rebumps trivially if another branch claims it first.


zero-bang (approved):
LGTM — clean test-only PR, no source changes.

**Verified:**
- Hand-traced all four asserted series against the two-pointer recency logic in `hours_since_big_move`/`hours_since_quantile_line` and ran the tests: all four pass. The cap-at-lookback (`min(idx - end, lookback)`), reset-to-0 at each line end, and lookback-fill (no-lines / empty-frame) cases are all exact.
- The architectural justification holds: `limen/features/__init__.py` omits the whole line-based family (`active_lines`, `active_quantile_count`, `quantile_line_density`), since they take `list[dict]` line structures rather than YAML-dispatchable scalars. Not registering these two mirrors that intentional pattern rather than leaving a gap.
- `run.py` updates the import block and the test list symmetrically; version + CHANGELOG bottom entry are correct.

**Engineering note:** `hours_since_big_move` accepts `start_idx` in its line dicts but only ever reads `end_idx` — the test correctly exercises this, and it's consistent with the documented signature, so no concern.

**Forward-looking:** the next predictable failure mode is someone later adding these to `features/__init__.py` or `docs/Features.md` assuming `func: limen.features.<name>` dispatch works — it would break at call time because the params are computed line lists, not scalars. The CHANGELOG entry documents the rationale, which mitigates this.

PR #589, opened 2026-06-08
Make the backtest execution model a swappable strategy

mikkokotila:
## What

Turns `backtest_snapshot`'s execution model into a swappable **strategy**, so the long/flat model we have today is one plug-in and new strategies (shorts, sizing, different fills) drop in later without touching the metric ledger.

- New `limen/backtest/long_flat_strategy.py` — the long-only, hold-while-1 fill-and-cost logic, lifted verbatim, exposed as `long_flat_strategy(predictions, open_px, close_px, price_change, *, execution_lag_bars, fee_bps, slip_bps) -> ExecutionResult`.
- `ExecutionResult` is a `NamedTuple(pos, gross, net)` of per-bar series — the single contract between a strategy and the ledger. Every one of the 20 columns flows from that triple.
- `backtest_snapshot` keeps its signature and output, gains a `strategy=` parameter (default `long_flat_strategy`), and shrinks to: **validate the price contract → delegate execution → build the ledger**.
- The `0/1` prediction validation and the `execution_lag_bars >= 0` guard move into the strategy that owns them (a future sized strategy will accept non-binary signals); their error messages keep their substrings, so existing tests are unaffected.

## Why this is the clean seam

The metric ledger already depended on exactly three series (`pos`, `gross`, `net`) — the function's own docstring called them "Atoms". The strategy follows the house dotted-path-function idiom (like indicators/targets/reference-architectures), not a new framework.

## Blast radius

- **Output is byte-identical** for the default path. `BACKTEST_SNAPSHOT_COLUMNS` is unchanged, so **downstream Furnace is unaffected** (no pin bump / Furnace issue needed).
- The three callers (`ReferenceModel._compute_backtest`, the rule-based backtest, `experiment_backtest_results`) use the default and are untouched.

## Scope

This is the **code seam** only — add a strategy by writing a function and passing `strategy=`. Wiring strategy selection through the manifest / YAML `backtest:` block (like `reference_architecture:`) is a deliberate follow-on once a second strategy exists.

## Test plan

- 3 new tests: strategy injection through `backtest_snapshot`, the `ExecutionResult` contract, and the negative-lag guard.
- Full local suite: **938/938 pass**; `ruff` clean; no circular import.
- Output parity covered by the existing snapshot tests (unchanged assertions).

Bumps to 3.24.0; CHANGELOG updated. Docs: `docs/Backtest.md` gains a "Pluggable strategy" section, package README updated.


limen/backtest/backtest_snapshot.py
```diff
@@ -139,61 +138,46 @@ def backtest_snapshot(df: pd.DataFrame,
...
+    gross = result.gross
+    net = result.net
+    pos = result.pos
     total_bars = len(df)
-    capital_fraction = pos.astype(float)
+
+    eq_net = (1.0 + net).cumprod()
     drawdown = (eq_net / eq_net.cummax().clip(lower=1.0)) - 1.0
+    in_market = pos > 0
+    entry_mask = in_market & (~in_market.shift(1, fill_value=False))
```
zero-bang, line 159: Non-blocking, forward-looking: the ledger now re-derives "trades" from 0→nonzero crossings of `pos`. For `long_flat_strategy` this is exactly its internal `entry_mask`, so `trades_per_bar` is correct here. But the PR positions a sizing strategy as the next plug-in, and a rebalance (e.g. `pos` 0.3→0.7→0.3) is a real fill that incurs cost yet keeps `in_market` True — so `trades_per_bar` would silently undercount, and this recomputed `entry_mask` would no longer match the fills the strategy actually charged for (which it knows internally but `ExecutionResult` doesn't expose). Worth deciding now whether `ExecutionResult` should carry turnover/fills so the ledger reads them rather than infers them. Fine to defer, but flag it in the contract docstring if so.

limen/backtest/backtest_snapshot.py
```diff
@@ -139,61 +138,46 @@ def backtest_snapshot(df: pd.DataFrame,
...
-    data['cvar_95_pnl_bps'] = _cvar_tail_bps(R_net)
+    data['wins_per_bar'] = round(float((net > 0).mean()), FRACTION_DECIMALS)
+    data['pnl_per_bar_bps'] = _mean_bps(net)
+    data['avg_win_bps'] = _mean_bps(net[net > 0])
+    data['avg_loss_bps'] = _mean_bps(net[net < 0])
+    data['cvar_95_pnl_bps'] = _cvar_tail_bps(net)
     data['trades_per_bar'] = round(float(entry_mask.sum()) / total_bars, RATE_DECIMALS)
-    data['inventory_per_bar'] = round(float(capital_fraction.mean()), FRACTION_DECIMALS)
-    data['cost_per_bar_bps'] = _mean_bps(R_gross - R_net)
+    data['inventory_per_bar'] = round(float(pos.mean()), FRACTION_DECIMALS)
```
zero-bang, line 179: Non-blocking contract note: `in_market = pos > 0` (line 158) and `inventory_per_bar = pos.mean()` bake long-only assumptions into the strategy-agnostic ledger. A future shorts strategy expressing `pos < 0` would be counted flat by `in_market` (breaking `trades_per_bar`/entry detection), and `inventory_per_bar` would net longs against shorts rather than measure gross exposure. If shorts are in scope for this seam, consider keying exposure off `pos.abs()` and documenting the sign convention on `ExecutionResult.pos`.

zero-bang (approved):
Clean, well-scoped refactor. I checked out the branch and verified the substance:

- **Output parity confirmed** — ran `backtest_snapshot` on a 200-bar random input against `main` and the full 20-column ledger is byte-identical (JSON-equal). The recomputed `entry_mask` (from `pos > 0` crossings) and `pos.mean()` reproduce the old `entry_mask` and `capital_fraction.mean()` exactly for the long-only path.
- **Callers safe** — all three (`base.py:161`, `rule_based.py:127`, `_experiment_backtest_results.py:95`) pass `execution_lag_bars=1` plus cost kwargs and rely on the default `strategy`; the new keyword-only param with a default is backward compatible.
- **Validation moved cleanly** — the `0/1` and negative-lag guards now live in `long_flat_strategy` with their error substrings preserved, and the price-contract checks stay in the ledger. `df.copy()` was dropped safely (the function no longer mutates `df`).
- Tests pass (19/19 backtest-related, including the 3 new ones), ruff clean, no circular import.

The `pos`/`gross`/`net` triple is a sound seam and the dotted-path-function idiom fits the house style. Two non-blocking notes on the `ExecutionResult` contract below — they don't affect this PR (the default path is exact) but are worth settling before the second strategy lands, since the ledger now infers trade/exposure semantics from `pos` rather than receiving them. Approving.

PR #593, opened 2026-06-08
Add a tunable notional_rate bet-size knob to the backtest

mikkokotila:
## What

Adds `notional_rate` — a tunable **bet size** — to the backtest economics, so a snapshot can be evaluated at the fraction of capital a live book actually deploys (a 10% book is `notional_rate=0.1`), not just all-in.

`notional_rate` is the fraction of capital deployed while in position, in `(0, 1]`, default `1.0` (all-in). `long_flat_strategy` scales its per-bar `pos`, `gross`, and `net` by it. The off-position capital sits in cash at 0, so the per-bar account return is exactly:

```
(1−k)·0 + k·[(1+r)·cost_mult − 1] = k · net_deployed
```

— a clean uniform scalar, no approximation.

## Effect on the ledger

- **Scales together** (`profit = edge − cost` still holds): `edge_bps`, `pnl_bps`, `cost_bps`, `drawdown_bps`, `pnl_per_bar_bps`, `avg_win/loss`, `cvar`, `cost_per_bar_bps`.
- **Invariant** (structure, not size): `wins_per_bar`, `trades_per_bar`.
- `inventory_per_bar` becomes the **average deployed notional** (`notional_rate` × share of bars in market).

So `notional_rate=1.0` is the pure signal view (today's behavior); dialing it down shows the account at that bet size.

## Wired through the YAML/CLI config, in lock-step with `fee_bps`/`slip_bps`

`BacktestConfig` → `Manifest.set_backtest_config(notional_rate=...)` → resolve/validate/inject → `ReferenceModel._cost_kwargs` → `backtest_snapshot`/`long_flat_strategy`, plus the YAML `backtest:` block, its compiler, the schema, and the `limen validate` rule:

```yaml
sfd:
  manifest:
    backtest:
      notional_rate: "{size}"   # swept bet size, in (0, 1]
  params:
    size: [0.1, 0.5, 1.0]
```

Validation rejects anything outside `(0, 1]` at the strategy, the manifest inject path, and `limen validate`.

## Blast radius

- **Default `1.0` keeps every existing backtest byte-identical** — the full suite passed unchanged before the new tests were added.
- `BACKTEST_SNAPSHOT_COLUMNS` is unchanged (no new columns; `notional_rate` scales existing ones), so **downstream Furnace is unaffected**.
- Architectures still don't expose cost/size params (asserted).

## Test plan

- 6 new tests: strategy scaling-vs-structure, `(0,1]` rejection at the strategy, manifest resolve/sweep + default, manifest rejection, and YAML valid/invalid/build.
- Full local suite: **944/944 pass**; `ruff` clean. Scaling verified numerically (×0.5 → all returns halve, inventory halves, wins/trades unchanged).

Bumps to 3.25.0; CHANGELOG and `docs/Backtest.md` updated.


docs/Backtest.md
```diff
@@ -41,25 +41,30 @@ The current snapshot backtest is intentionally simple and opinionated:
...
 
-In a YAML/CLI manifest the same configuration is a `backtest:` block under `sfd.manifest`, sibling to `target:` and `scaler:`. Each cost is a fixed number or a `"{param}"` reference into `sfd.params`, mirroring how indicator and target params are swept:
+`set_backtest_config` also takes `notional_rate` — the fraction of capital deployed while in position, in `(0, 1]` (default `1.0`, all-in; a 10% book is `notional_rate=0.1`). It scales `edge`, `pnl`, `cost`, and `drawdown` together so the ledger reflects the account at that bet size, leaves `wins_per_bar` and `trades_per_bar` untouched, and turns `inventory_per_bar` into the average deployed notional. Like the costs it is a fixed number or a search-param name, so one search can sweep bet size.
```
zero-bang: `drawdown` does **not** scale linearly with `notional_rate` — unlike `edge`/`pnl`/`cost`, drawdown is a path-dependent compounding metric (`(1+k·net).cumprod()` against its running peak), so `k` does not factor out. Verified numerically against this branch on a path with consecutive losing bars:

```
drawdown_bps_p5   full=-5000.0  half(k=0.5)=-2777.8  0.5*full=-2500.0  -> NOT linear
drawdown_bps_p50  full=-4166.7  half(k=0.5)=-2222.2  0.5*full=-2083.4  -> NOT linear
pnl_per_bar_bps   full=-830.2   half=-415.1          0.5*full=-415.1   -> linear
edge_bps_p50      full=-1428.6  half=-714.3          0.5*full=-714.3   -> linear
```

The deviation is risk-relevant: at half size the worst-5% drawdown is ~11% *deeper* than a linear halving would suggest, so a reader trusting "scales drawdown together" to reason about risk at a smaller book is misled toward understating tail risk. Note your own scaling test (`test_backtest_snapshot_notional_rate_scales_returns_not_structure`) checks `pnl_per_bar_bps`, `edge_bps_p50`, and `inventory_per_bar` but deliberately omits `drawdown_bps` — consistent with this.

Fix: drop `drawdown` from the "scales ... together" list here (and the matching line 44 "scales every return column" and the CHANGELOG v3.25.0 entry), or qualify it — e.g. "scales `edge`, `pnl`, and `cost` together; `drawdown` moves with bet size but path-dependently, not as a clean scalar."

zero-bang (commented):
Solid, well-scoped change. The economic core is right: computing `net` on the unscaled position and then multiplying `pos`/`gross`/`net` by `notional_rate` yields exactly the cash-plus-deployed account return `(1−k)·0 + k·net_deployed`, so the sign-based columns (`wins_per_bar`, `trades_per_bar`) stay invariant and the linear ones scale cleanly. The config threading is in genuine lock-step with `fee_bps`/`slip_bps` across the manifest, compiler, schema, and validation rule, and validation rejects out-of-range values at all three surfaces (strategy, inject path, `limen validate`). The default-1.0 byte-identity argument holds and the injected-strategy test uses `**kwargs`, so the new required kwarg doesn't break custom strategies. Ran the affected suites locally (185 passed).

One accuracy issue on the `drawdown` scaling claim (inline) — the code is correct, but the documentation overstates linearity for one column. Submitting as a comment rather than a block since it's a doc-wording fix, not a code defect.

Forward-looking: the next predictable failure mode is someone reading two snapshots run at different `notional_rate` and treating the `drawdown_bps_*` ratio as the bet-size ratio to back out an implied size — that inference is wrong because drawdown is path-dependent, and it errs toward *understating* tail risk at smaller sizes (at k=0.5 the worst-5% drawdown came in ~11% deeper than a linear halving predicts).

mikkokotila:
Thanks — addressed all four points in 4e79149:

1. **`notional_rate` validation too loose (Copilot, `long_flat_strategy.py`)** — fixed. The strategy now rejects `bool` and non-numeric values with a clear `ValueError` (`isinstance(notional_rate, bool) or not isinstance(notional_rate, numbers.Real) or not 0 < notional_rate <= 1`), matching the manifest/YAML surfaces. Added `True`/`'half'` to the rejection test.

2. **`drawdown` does not scale linearly (zero-bang)** — correct and the most important catch; fixed. Dropped `drawdown` from the "scales together" claim in `docs/Backtest.md` (the assumptions bullet and the detail paragraph) and the v3.25.0 CHANGELOG entry, and qualified it: `edge`/`pnl`/`cost` scale together; `drawdown` moves with bet size path-dependently (it compounds against a running peak, so the factor doesn't cancel). The scaling test already (correctly) omitted `drawdown_bps`.

3. **Misleading bool error message (Copilot, `rules.py`)** — fixed. The YAML bool branch is now bound-accurate per key: `notional_rate` reports "a number in (0, 1]" with a `0.1` example; fee/slip keep "a non-negative number".

4. **Always forwarding `notional_rate` is a contract change (Copilot, `backtest_snapshot.py`)** — intentional, now documented. `backtest_snapshot` forwards the full execution kwarg set, and that set grows as the backtest gains knobs; I added a note to the Pluggable-strategy docs that strategies should accept those keywords or absorb extras with `**kwargs` (as the injected-strategy test does). I kept the unconditional forward rather than adding signature inspection — the seam is new with no external consumers, and inspection would add complexity against the simplicity mandate.

944/944 tests pass, ruff clean.

zero-bang (approved):

mikkokotila:
Second round (51a04ae) — addressed the remaining comments:

**Breaking-change / always-forwarding `notional_rate` (Copilot, `backtest_snapshot.py` + the CHANGELOG compat note)** — fixed structurally rather than documented around. On reflection the comment pointed at a real design smell: `notional_rate` is a *uniform scalar on the output*, not a fill-mechanics concern like `fee_bps`/`slip_bps`, so it doesn't belong in the strategy signature. `backtest_snapshot` now applies it as a post-scale on the returned `(pos, gross, net)`, and `long_flat_strategy` reverts to its exact v3.24.0 signature. So the pluggable-strategy contract is unchanged and existing custom strategies keep working — no `TypeError`, no compat impact to call out. Output is identical (verified: ×0.5 → returns and inventory halve, wins/trades invariant). The validation/range test moved to the `backtest_snapshot` level accordingly.

**`BacktestCostSpec` docstring grammar (Copilot, `rules.py:1008`)** — fixed ("or any a" → clean rephrase).

**`{param}` placeholder inconsistency (Copilot, `rules.py:1033` & `:1042`)** — fixed. The bool-branch and invalid-ref suggestions now use a single generic `{my_param}` placeholder instead of the mixed `{fee}`/`{p}`.

944/944 tests pass, ruff clean. The net effect is a *smaller* diff — `long_flat_strategy.py` is now untouched relative to main.

PR #601, opened 2026-06-10
Fix cusum_filter lazy path, logreg n_jobs overhead, and cache HF dataset snapshots

mikkokotila:
## What

Three fixes surfaced by a comprehensive performance-profiling campaign on `limen run` (14 manifest variants, ~270 measured rounds, full cProfile):

1. **`cusum_filter` lazy compatibility** — the feature subscripted the frame (`data[close_col]`), which raises `LazyFrame is not subscriptable` inside `_apply_feature_transforms` (the manifest pipeline pipes LazyFrames). Any YAML manifest listing the feature crashed its entire run on round 0. Column access is now expression-based (`pl.col(...).map_batches(...)`), mirroring `fractional_diff`; eager output is byte-identical, with a LazyFrame/eager parity test added.

2. **`logreg_binary` `n_jobs` default `-1` → `None`** (and `[None]` in the foundational SFD preset). Binary classification fits one class, so `-1` routes a single task through a sleep-polled joblib/loky pool. Measured on real prepared manifest data (5,393×41, M1 Max, sklearn 1.7.2): lbfgs fit median **1,072 ms with `-1` vs 57 ms with `None` (19×)**; cProfile of an 8-round run showed 4.67 s of 8.54 s total inside `time.sleep` under `LogisticRegression.fit`. liblinear ignores the knob (~30 ms either way). Explicit `n_jobs` in manifests keeps working.

3. **Disk cache for Hugging Face dataset snapshots** at `~/.cache/limen/datasets/{org}--{repo}/{file_name}`. Every fetch previously re-downloaded the full snapshot (1.3–3.2 s native; 60 s / 452 MB on the 1M-source aggregation path), and one mid-body CDN read timeout aborted the whole experiment — observed 3 times in ~20 fetches while profiling; same root cause as the recurring CI 429 flake. `latest.json` still resolves fresh every run and snapshot file names are immutable and date-stamped, so `repo + file_name` is an exact key with **no staleness path**: a new snapshot publishes a new name → cache miss → atomic write (unique temp + rename) → superseded snapshots pruned. Non-HF URLs and local files are never cached. Covered by cache-hit/prune and non-HF passthrough tests.

## Validation

- `ruff` clean on all changed files
- Targeted suites green: `test_historical_data` + `test_feature_library_primitives` (44), `test_e2e_loop_trainer_sensor` + `test_manifest_prepare_data` + `test_yaml` (151)
- Full `tests.run` suite + end-to-end manifest verification runs included in PR conversation


limen/data/historical_data.py
```diff
@@ -238,6 +242,73 @@ def _read_remote_bytes(url: str) -> bytes:
...
+
+    if not url.startswith(_HUGGINGFACE_DATASETS_URL_PREFIX):
+        return None
+
+    path_parts = [part for part in urlparse(url).path.split('/') if part]
+    if len(path_parts) < _HUGGINGFACE_DATASET_REPO_PART_COUNT + 1 or path_parts[0] != 'datasets':
+        return None
+
+    repo_dir = f'{path_parts[1]}--{path_parts[2]}'
+    return _DATASET_CACHE_DIR / repo_dir / path_parts[-1]
```
zero-bang, line 282: Non-blocking: this cache key (`org--repo / file_name`) is only collision-safe as long as snapshot file names are immutable per content. The correctness rests entirely on that upstream convention — there's no content check (size/hash/etag) on a cache hit. If a snapshot is ever re-uploaded under a reused name (corrected re-upload, or a future switch to a fixed filename like `latest.parquet`), this would serve stale bytes indefinitely with no recovery path. Fine to ship given today's date-stamped naming, but worth documenting as a load-bearing invariant (or guarding with an etag check) so a future upstream change doesn't silently poison the cache.

limen/data/historical_data.py
```diff
@@ -238,6 +242,73 @@ def _read_remote_bytes(url: str) -> bytes:
...
+    if cache_path is None:
+        return _read_remote_bytes(url)
+
+    if cache_path.exists():
+        return cache_path.read_bytes()
+
+    content = _read_remote_bytes(url)
+    cache_path.parent.mkdir(parents=True, exist_ok=True)
+    tmp_path = cache_path.with_name(
+        f'{cache_path.name}.{os.getpid()}{_DATASET_CACHE_TMP_SUFFIX}'
```
zero-bang: Minor robustness: the temp name is keyed on `os.getpid()` only, so two threads in the *same* process that both miss the cache for the same snapshot would share a temp path — the second `tmp_path.replace(cache_path)` can then raise `FileNotFoundError` after the first already moved it. Data prep is single-fetch/serial today so this won't bite in practice, but a `uuid4().hex` (or `threading.get_ident()`) suffix alongside the pid would make the unique-temp guarantee actually unique. Non-blocking.

zero-bang (approved):
LGTM — three well-scoped, well-justified fixes.

**Verified:**
- `cusum_filter`: the `pl.col(...).cast(pl.Float64).map_batches(...)` rewrite is a faithful mirror of `fractional_diff` (limen/features/fractional_diff.py:105), and `manifest_core` collects with a plain `.collect()` (limen/experiment/manifest_core.py:1072 et al.), not the streaming engine — so the stateful `_cusum_events` receives the full series in one batch and output is byte-identical to the eager path. The lazy/eager parity test confirms it. Good catch on the original `data[close_col]` subscript crashing the whole run.
- `n_jobs -1 → None`: this only changes joblib dispatch, not model output, so SFD search reproducibility is untouched while removing the sleep-polled pool overhead. Both the reference arch and the foundational preset are updated consistently.
- Disk cache: full content is fetched before the write, and the write is temp-file + atomic `replace`, so a mid-body timeout (the failure you were chasing) never leaves a partial file cached. Prune keeps exactly one file per repo, so the cache is self-bounding. Non-HF passthrough is tested.

**Engineering note:** the entire "no staleness path" guarantee rests on an upstream convention (HF snapshot file names being immutable and date-stamped). That's a sound bet today, but it's a load-bearing external assumption — see the inline note. Two non-blocking observations inline; neither blocks merge.

**Checks:** Ruff green. Tests/CodeQL/Workers pending at review time — none failing, and I see no security surface or logic gap that would flip them red. Approving on that basis.

**Forward-looking:** the next predictable failure mode is upstream naming drift — if any `vaquum/*` dataset ever republishes under a *reused* file name (e.g. a corrected re-upload of the same date-stamped snapshot, or a switch to a fixed `latest.parquet`), the `repo + file_name` key would hit the cache and serve stale bytes forever, with no validation to catch it. Worth a one-line guard or a documented invariant on the dataset-publishing side.

mikkokotila:
## End-to-end verification (identical manifest, identical seeded param draws)

12-round manifest run, 7,119 rows, before (v3.27.0) vs after (this branch):

| | sum exec (12 rounds) | mean round | lbfgs rounds | liblinear rounds | wall |
|---|---|---|---|---|---|
| before | 7.70 s | 0.642 s | 0.740 s | 0.150 s | 10.85 s |
| after, cold cache | 1.65 s | 0.138 s | 0.138 s | 0.135 s | 4.96 s |
| after, warm cache | 1.60 s | 0.133 s | 0.134 s | 0.130 s | 3.43 s |

**4.8× per round**; the lbfgs/liblinear gap collapses to noise (the gap was the joblib dispatch). A manifest including `cusum_filter` now completes end-to-end (previously crashed on round 0). Full `tests.run` suite green locally (exit 0); the suite's own real HF fetches populated the new cache as a side effect, confirming the cache path in practice.

## Reproducibility audit

- `liblinear` rounds: byte-identical metrics before vs after.
- `lbfgs` rounds: metrics shift ~±0.003 AUC (tp/fp counts ±8 of ~250). Verified cause: the worker pool changed the floating-point execution path, and at `tol=0.001` lbfgs stops within tolerance of a slightly different optimum — same-config fits are bit-identical run-to-run (max coef diff 0.0), and the old-vs-new coefficient gap shrinks ~8.5× (0.0173 → 0.0020 on a 2.4 coef scale) when `tol` tightens to 1e-8. Old `n_jobs=-1` results were already pool- and machine-dependent in the same way. Noted in the CHANGELOG.

limen/data/historical_data.py
mikkokotila: Documented in ed0378a: the docstring now states the immutable-file-name convention as a load-bearing invariant on the Vaquum dataset-publishing side, that there is no content check on a hit, and the recovery path (delete the cached file). Went with documentation over an etag HEAD per hit — that would reintroduce a per-run network roundtrip and the flake surface this cache exists to remove.

limen/data/historical_data.py
mikkokotila: Fixed in ed0378a — temp names now carry pid + uuid4, so same-process concurrent misses get genuinely unique temps.

limen/data/historical_data.py
```diff
@@ -238,6 +242,73 @@ def _read_remote_bytes(url: str) -> bytes:
...
+    )
+    tmp_path.write_bytes(content)
+    tmp_path.replace(cache_path)
+
+    for stale in cache_path.parent.iterdir():
+        if stale != cache_path and not stale.name.endswith(_DATASET_CACHE_TMP_SUFFIX):
+            stale.unlink(missing_ok=True)
+
+    return content
```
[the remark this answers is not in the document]
mikkokotila: Done in ed0378a — cache reads and writes are wrapped in OSError handling that logs a warning and falls back to the direct download (a cache failure never fails a fetch that succeeded remotely), and pruning now skips non-files. Covered by test_read_any_file_falls_back_to_download_when_cache_is_unwritable.

limen/data/historical_data.py
```diff
@@ -238,6 +247,100 @@ def _read_remote_bytes(url: str) -> bytes:
...
+                and not stale.name.endswith(_DATASET_CACHE_TMP_SUFFIX)
+            ):
+                stale.unlink(missing_ok=True)
+    except OSError as exc:
+        logger.warning(
+            'Dataset cache write failed for %s (%s) — proceeding uncached',
+            cache_path, exc,
+        )
+
+    return content
```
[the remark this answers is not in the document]
mikkokotila: Done in e8b02d9 — the prune loop now has its own OSError handler with an accurate message ('the cached snapshot is intact'); 'proceeding uncached' is only reported when the write/replace itself fails.

limen/features/cusum_filter.py
```diff
@@ -28,14 +28,22 @@ def cusum_filter(
     Returns:
         pl.DataFrame: The input data with an Int8 column: 1 for an up event,
             -1 for a down event, 0 otherwise
+
+    NOTE: Column access is expression-based so the feature runs on both eager
+    DataFrames and the LazyFrames piped by the manifest feature pipeline.
```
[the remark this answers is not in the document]
mikkokotila: Done in e8b02d9 — Args/Returns now document pl.DataFrame | pl.LazyFrame with the return type matching the input frame type.

PR #610, opened 2026-06-14
docs: complete canonical docs coverage

mikkokotila:
## Summary
- add missing canonical CLI and package README surfaces
- route all public docs and package READMEs through the docs-site assembly map
- expand docs coverage for CLI, transforms, release/version local scope, technical debt rules, and package boundaries
- bump package version to 3.30.1 and add changelog entry

## Review gates
- Phase 1 coverage: APPROVED by codex exec gpt-5.5 xhigh
- Phase 2 comprehensiveness: APPROVED by codex exec gpt-5.5 xhigh
- Phase 3 consistency/coherence: APPROVED by codex exec gpt-5.5 xhigh

## Local checks
- docs-site npm run check
- git diff --cached --check


zero-bang (approved):
LGTM — thorough, accurate docs-only PR. I verified the substantive claims against source rather than against other docs:

**Verified**
- All 10 newly-mapped `source:` paths in `assemble-docs.mjs` exist and the assemble step runs clean (it `readFile`s every source, so a broken mapping would fail). The local `docusaurus build` failure in `npm run check` is only a missing binary in my environment, not a content issue.
- **Complete coverage confirmed:** all 16 `limen/*/README.md` files are now routed through the assembly map (1:1, no orphans), and zero `docs/**/*.md` pages remain unmapped. The PR fully delivers its stated "route all public docs and package READMEs" goal.
- `Command-Line-Interface.md` and `cli/README.md`: command set, options (`--parent`, `-m/--message`, `--dry-run`, `--resume`, `--template`, `--backup-remote`), and the `manifest://` run path all match `limen/cli/main.py`.
- The repeated "`profile` is static because YAML rejects `test_data_source`" claim is correct: `yaml/rules.py:353` rejects `sfd.manifest.test_data_source`, and `profiler.py:134` skips runtime sampling when `test_data_source_config is None`. The doc correctly reconciles this with the general profiler docstring.
- `calibration/README.md` entry points and the `fit_calibrator` / `apply_calibrated_predict` return shapes (`_preds`, `_probs`, `optimal_threshold`, `val_score`) match `pipeline.py` exactly.
- `targets/README.md` class list matches `targets/__init__.py` (all 15 exports). Transform signatures in `Transforms.md` match `limen/transforms/*`.
- Local link rewrites in `docs/README.md` (Making-Release, Semantic-Versioning) point to files that exist.
- CHANGELOG entry + `pyproject.toml` bump to `3.30.1` are present and consistent.

**Open threads**
- The Copilot bot comment is a quota notice, not an objection — nothing to address.
- CI: Ruff passed; Tests/CodeQL/Workers-build are pending. No Python source changed (docs + assemble `.mjs` + version only), so these are low-risk; the docs-site assemble step already passed locally.

**Forward-looking carry-forward (non-blocking):** the assembly map now has full README coverage, but nothing *gates* it — a future package that adds a `README.md` without a map entry would be silently orphaned and the build would still pass. A small CI assertion that every `limen/*/README.md` appears in `assemble-docs.mjs` would lock in the completeness this PR just achieved.

zero-bang (approved):

PR #612, opened 2026-06-14
docs: align docs with honesty rules

mikkokotila:
## Summary
- remove second-person voice, rhetorical questions, vague praise, and unsupported approximation markers across canonical docs and package READMEs
- replace fake-looking ellipses and placeholder snippets with concrete examples or explicit prose
- align headings and routing copy with the Vaquum voice addendum while preserving the existing docs structure

## Checks
- content scan: 0 second-person, vague-praise, approximation, rhetorical punctuation, placeholder, and ellipsis hits
- git diff --check
- npm run check

### DIFF
--- a/docs/Backtest.md
+++ b/docs/Backtest.md
@@ -100,9 +100,19 @@ from limen.backtest.long_flat_strategy import ExecutionResult
 
 def my_strategy(predictions, open_px, close_px, price_change, *,
                 execution_lag_bars, fee_bps, slip_bps) -> ExecutionResult:
-    position = predictions.astype(float)
+    position = predictions.astype(float).shift(execution_lag_bars, fill_value=0.0)
     gross_return = position * price_change / open_px
-    net_return = gross_return
+
+    entry_mask = (position == 1.0) & (position.shift(1, fill_value=0.0) == 0.0)
+    exit_mask = (position == 1.0) & (position.shift(-1, fill_value=0.0) == 0.0)
+    fee = fee_bps / 10_000.0
+    slip = slip_bps / 10_000.0
+    cost_mult = position.copy()
+    cost_mult[:] = 1.0
+    cost_mult.loc[entry_mask] *= (1.0 - fee) / (1.0 + slip)
+    cost_mult.loc[exit_mask] *= (1.0 - fee) * (1.0 - slip)
+
+    net_return = ((1.0 + gross_return) * cost_mult) - 1.0
     return ExecutionResult(pos=position, gross=gross_return, net=net_return)
 ```
 
--- a/docs/Backtest.md
+++ b/docs/Backtest.md
@@ -2,13 +2,13 @@
 
 Backtest is Limen's trading-economics layer. It takes prediction outputs and asks the next question after benchmark:
 
-if we traded this signal as a simple long-only strategy, what would the return profile look like after costs?
+The layer maps a binary signal into long-only returns after fees and slippage.
 
 Limen exposes a vectorized snapshot backtest, used throughout the `Log` layer.
 
-## Where Backtest Lives
+## Where backtest lives
 
-The most common backtest outputs are:
+Backtest outputs are exposed through:
 
 - `uel.experiment_backtest_results`
 - `uel._log.experiment_backtest_results()`
@@ -28,7 +28,7 @@ and returns one summary row.
 
 ### Current assumptions
 
-The current snapshot backtest is intentionally simple and opinionated:
+The current snapshot backtest has a fixed long-flat contract:
 
 - long-only
 - direct snapshot predictions must already be binary `0/1`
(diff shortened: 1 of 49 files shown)


docs/Experiment-Manifest.md
```diff
@@ -637,11 +643,11 @@ from limen.sfd.reference_architecture import logreg_binary
 
 The model function must accept the prepared `data_dict` as its first argument.
 
-All remaining parameters are auto-mapped from `round_params` by signature inspection. For example, if your model function has:
+All remaining parameters are auto-mapped from `round_params` by signature inspection. For a model function with this signature:
 
 ```python
 def logreg_binary(data, C=1.0, class_weight=None, max_iter=100, solver='lbfgs'):
-    ...
+    return LogRegBinary().evaluate(data, inline_metrics=True, C=C, class_weight=class_weight, max_iter=max_iter, solver=solver)
```
zero-bang: This concrete example is not runnable as written. `LogRegBinary.evaluate` has the signature `evaluate(self, data, inline_metrics=True)` (`limen/sfd/reference_architecture/logreg_binary.py:116`) — it does not accept `C`, `class_weight`, `max_iter`, or `solver`, so this line raises `TypeError: evaluate() got an unexpected keyword argument 'C'`. The hyperparameters go to `train()`; the real `logreg_binary()` function (`logreg_binary.py:194-213`) does `model = LogRegBinary(...).train(data, C=C, ...)` then `model.evaluate(data, inline_metrics=True)`.

Suggested fix (note `train()` returns the instance):
```python
def logreg_binary(data, C=1.0, class_weight=None, max_iter=100, solver='lbfgs'):
    model = LogRegBinary().train(data, C=C, class_weight=class_weight, max_iter=max_iter, solver=solver)
    return model.evaluate(data, inline_metrics=True)
```
The old `...` placeholder sidestepped this; the replacement should stay code-true since this page's whole point is signature-based auto-mapping.

docs/Backtest.md
```diff
@@ -100,8 +100,10 @@ from limen.backtest.long_flat_strategy import ExecutionResult
 
 def my_strategy(predictions, open_px, close_px, price_change, *,
                 execution_lag_bars, fee_bps, slip_bps) -> ExecutionResult:
-    ...
-    return ExecutionResult(pos=..., gross=..., net=...)
+    position = predictions.astype(float)
+    gross_return = position * price_change / open_px
+    net_return = gross_return
```
zero-bang: Non-blocking: `net_return = gross_return` discards the `fee_bps`/`slip_bps` args the function receives, yet `ExecutionResult.net` is documented as "per-bar net return after entry and exit costs" (`long_flat_strategy.py:20`, and Backtest.md line 109). Because `backtest_snapshot` only scales the returned triple by `notional_rate` and never re-applies costs, a custom strategy that returns `net == gross` silently produces a cost-free `net` ledger. Either apply the costs in the example (mirroring `long_flat_strategy.py:96-105`) or add a one-line comment like `# this minimal strategy models no costs` so the unused `fee_bps`/`slip_bps` don't read as an oversight.

zero-bang (changes requested):
Thorough, well-scoped voice/honesty pass — removing the unsupported "1,000 papers / zero breaking bugs" claim, the second-person voice, and the vague-praise adjectives genuinely improves the corpus. I verified every place where the PR replaced a `...`/placeholder with concrete code against the actual source. Findings:

**Verified accurate (good):**
- `docs/Backtest.md` fee/slip `5.0` bps per fill, `20` bps entry+exit, and `notional_rate` default `1.0` applied post-strategy — matches `manifest_core.py` `BacktestConfig` and `backtest_snapshot.py:165-167`.
- `docs/Calibration.md` expanded manifest chains, imports, and the `passthrough_calibrator` / `fixed_threshold_optimizer` protocol examples — all real and correctly shaped.
- `docs/Developer/Pruning-Strategies.md` op list (`inject_value`) and `keep_is` — both real (`feedback_controller.py:313-324`).
- `docs/Experiment-Manifest.md` rule-based `data_dict` condition keys and `_alignment` keys — match `manifest_core.py:1945` and `splits.py:126-130`.
- The backtest custom-strategy `gross_return = position * price_change / open_px` is correct because `price_change` is `close - open` (absolute), matching `long_flat_strategy.py:90`.

**One blocking issue:** the new concrete `logreg_binary` model-function body calls `.evaluate()` with model hyperparameters it does not accept — copied verbatim it raises `TypeError`. Inline comment below.

**Engineering note / carry-forward:** the PR's content scan checks voice/style but does not execute doc code blocks, so example-vs-source drift like the `evaluate()` issue isn't gated by CI. Worth a future doctest or import-check on fenced Python in `docs/`. Next predictable failure mode: a later refactor of `LogRegBinary.train`/`evaluate` signatures silently invalidates these now-concrete examples with nothing to catch it.

CI status: Ruff passed (relevant for a docs-only change); Tests/CodeQL/Workers pending, none failing.

zero-bang (approved):

PR #614, opened 2026-06-15
fix: sensor reproducibility via context carry-over and strict null detection

### LINKED ISSUE #120: feat: move confidence gate into metrics libraries
regime[conf < CONFIDENCE_THRESHOLD] = 0

### LINKED ISSUE #598: Served predictions diverge ~5% on test: causal_rolling_robust scaled per-split in training but continuously in serving
## Background

Validating the conduit prediction cache end-to-end (Limen 3.27.0; a `dollar_60M` toxicity-gate experiment — logreg + calibration), I compared a served cohort's predictions against the experiment's own saved predictions on the **test** window. This became possible after the predict-guard switches (#590 / #591 / #592) let val/test serve real predictions instead of being masked.

Single-permutation cohort (so the served prediction equals one sensor); permutation `ba76bffb8896e19787b16294c3ed6e7d11579b79fd547ba1724ea96ee95fad12` of experiment `toxicity_gate_validate` round 0002; `scaler_type: causal_rolling_robust`.

## Findings

On the test window (1,472 bars) the served predictions agree with the experiment's saved predictions **94.84% (1,396 / 1,472)** — 76 bars flip. The flips are not boundary ties (55% are more than 0.05 from the decision threshold; model-probability gaps up to 0.23). The gap is reproducible: reconstructing the sensor twice gives 100% identical predictions. A backtest of the served signal is correspondingly close but not exact.

## Diagnosis

The gap is **not** in the data fed to serving, and **not** in the model or calibration. It is in how the `causal_rolling_robust` scaler is applied: **per-split in training, but over the continuous series in serving.**

Evidence, each isolating one layer:

1. **The data range is not the cause.** Running the real serving path (`predict_all`) over the date-limited data the experiment used (30,781 rows, 2024-01-01 → 2026-05-15) and over the full arrow (91,407 rows, 2020 → now) both give **1,396 / 1,472 = 94.84%** versus the experiment. The data window changes nothing.
2. **Model and calibration are faithful.** Running the serving model on the experiment's own batch `x_test` reproduces the experiment's predictions **1,472 / 1,472 = 100%**. The gap is upstream of the model.
3. **The features diverge with a sharp fingerprint.** Comparing online features (`sensor_input_prep`) to batch features (`prepare_data`) per test bar: bars 0–999 differ, and **every bar from 1000 onward is byte-identical**. All 76 prediction flips fall on differing bars. 1000 = the scaler's window length.

Mechanism — `CausalRollingRobustScaler` (`limen/scalers/causal_rolling_robust_scaler.py`) is a **1000-row trailing-window** median/IQR, with a fallback to the train-fitted median/IQR for rows that have fewer than `min_samples` (50) trailing observations:

- `MLManifest.prepare_data` (`manifest_core.py`, the `for i, split in enumerate(split_data)` loop) calls `_apply_scaler` on **each split in isolation**. `transform()` on the test split therefore sees only test rows: the 1000-row window **restarts at the test boundary**, so the first ~1000 test rows are scaled cold (the first 50 on the train fallback, then a partial window).
- `MLManifest.sensor_input_prep` (docstring: "Prepare raw klines for live inference **without splitting**") calls `_apply_scaler` once on the **full continuous series**. `transform()` on a test bar sees the preceding val/train rows, so the 1000-row window is **fully warm**.

Same scaler and same fitted params, but a different row context — cold per-split in training, warm continuous in serving. The first 1000 test bars are scaled differently and everything past the window length is identical, which is exactly the measured fingerprint.

---
*Supersedes #597, which described the direction backwards (it had serving as the cold side; training is the cold side).*

### LINKED ISSUE #599: prepare_data drop_nulls() removes any null-containing row regardless of position, not just warm-up
## What

`MLManifest.prepare_data` drops rows twice inside its per-split loop (`limen/experiment/manifest_core.py:1084` and `:1086`):

```python
for i, split in enumerate(split_data):                          # 1069
    lazy = _apply_feature_transforms(self, lazy, round_params)  # 1071  features / indicators
    data = lazy.collect()
    ...
    data = data.fill_nan(None).drop_nulls()                     # 1084
    data, all_fitted_params = _apply_scaler(...)                # 1085
    split_data[i] = data.fill_nan(None).drop_nulls()            # 1086
```

`drop_nulls()` has no "only the start" rule — it removes every row that has a null in **any** column, wherever that row sits. `fill_nan(None)` runs first, so NaNs count as nulls too.

## In practice

The nulls cluster at the **start** of each split (rolling indicators are null until their window fills), so most of what's dropped is the leading warm-up block. But it is **not** restricted to the start: any null that appears mid-split — a gap in the data, or a feature/indicator that goes NaN on some bar — drops that row in place.

## Consequence

Rows can be silently removed from the middle of train/val/test. Downstream row-windowed steps count the **row index, not the timestamp** — e.g. the `causal_rolling_robust` trailing-1000-**row** window (#598) — so a mid-split drop shifts those windows relative to the continuous live stream, where no such drop happens.

pdey:

## What does this PR change?

- Fix rolling scaler reproducibility: the causal rolling scaler cold-started at each split boundary during training, causing val and test bars to receive different scaled values than the continuous inference stream. Context Carry-Over prepends the raw tail of the preceding split before the feature transform pass and strips it after scaling, so the scaler is always warm when it reaches the actual split data. The inference path is unchanged.
- Expose a context rows property on the causal rolling robust scaler so CCO can detect the required warm-up size automatically. Stateless scalers return zero via duck-typing and receive no CCO block.
- Add strict mode to the ML manifest. When enabled, unexpected nulls remaining after CCO dislodgement raise an error rather than logging a warning. The experiment loop catches the error per permutation, records the failure in results, and continues to the next round.
- Add constructor parameter support for scalers in both the manifest API and YAML. Parameters such as window size and minimum samples can be set as literals or as sweep-param references, letting them appear in results and be fully reproducible.
- Add a data quality section to the profile output. The profiler captures null warnings during sample rounds and the CLI renders them under a dedicated header.

Closes #598
Closes  #599 

## Checklist

- [x] If this PR delivers a PRD’s **last** remaining slice, I also linked the parent PRD to close with it (e.g., “Closes #120”), so the PRD closes on merge instead of by hand

### DIFF
--- a/limen/experiment/experiment_core.py
+++ b/limen/experiment/experiment_core.py
@@ -526,6 +526,8 @@ class UniversalExperimentLoop:
                 if caught else '[]'
             )
 
+            round_succeeded = '_alignment' in data_dict
+
             if 'extras' in round_results:
                 extras = round_results.pop('extras')
                 if post_processing:
@@ -541,12 +543,12 @@ class UniversalExperimentLoop:
             current_preds = round_results.pop('_preds', None)
             if current_preds is None:
                 current_preds = []
-            if post_processing and '_alignment' in data_dict:
+            if post_processing and round_succeeded:
                 self.preds.append(current_preds)
-            if post_processing and '_alignment' in data_dict and '_scaler' in data_dict:
+            if post_processing and round_succeeded and '_scaler' in data_dict:
                 self.scalers.append(data_dict['_scaler'])
 
-            if post_processing and '_alignment' in data_dict:
+            if post_processing and round_succeeded:
                 self._alignment.append(data_dict['_alignment'])
 
             round_results['id'] = current_hash
@@ -554,7 +556,7 @@ class UniversalExperimentLoop:
                 time.time() - start_time, 2,
             )
 
-            if post_processing and '_alignment' in data_dict:
+            if post_processing and round_succeeded:
                 self.round_params.append(sfd_params)
             for key, value in round_params.items():
                 round_results[key] = value
@@ -581,7 +583,7 @@ class UniversalExperimentLoop:
                         current_round, sorted(extra_keys),
                     )
 
-            if round_data_path and '_alignment' in data_dict:
+            if round_data_path and round_succeeded:
                 self._append_round_data(
                     round_data_path, current_hash, sfd_params,
                     current_preds,
--- a/limen/experiment/experiment_core.py
+++ b/limen/experiment/experiment_core.py
@@ -228,6 +228,9 @@ class UniversalExperimentLoop:
                     f'UniversalExperimentLoop Existing results CSV has no header: {csv_path}'
                 )
 
+        data_dict: dict[str, Any] = {}
+        _pending_csv_rows: list[dict] = []
+
         for i in tqdm(range(n_permutations)):
 
             # Start counting execution_time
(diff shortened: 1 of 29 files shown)


limen/experiment/experiment_core.py
```diff
@@ -489,13 +490,18 @@ def _run_with_msq(self,
...
             try:
                 with warnings.catch_warnings(record=True) as caught:
                     warnings.simplefilter('always')
                     data_dict = self.prep(self.data, round_params=sfd_params)
                     round_results = self.model(
                         data=data_dict, round_params=sfd_params,
                     )
+            except StrictModeError as exc:
+                logger.error('Round %d failed strict mode check: %s', current_round, exc)
+                round_results = {'strict_mode_error': str(exc)}
```
zero-bang, line 534: Blocking: a strict-mode failure on the **first** permutation corrupts the `results.csv` schema for the whole run. On failure `round_results` is this minimal dict (after the later additions it holds only `strict_mode_error`, `_warnings`, `id`, `execution_time`, and the param columns — no metric columns). If this is the first row, `write_header` is True and `csv_header = list(round_results.keys())` (line 575) is frozen without `accuracy`/`auc`/etc. Every subsequent **successful** round then hits `extra_keys` and its metrics are silently dropped from the CSV (only a `logger.warning`). `Cohort` reads `results.csv` and selects on `accuracy`, so it would break or mis-select. Note the in-memory `self.experiment_log` survives this (it uses `vertical_relaxed`), which is why the existing test — where round 2, not round 1, fails — doesn't catch it.

Since `strict_mode=True` is now the default across all foundational SFDs/templates and is meant to fire exactly on data-quality edge cases, a first-round failure is plausible. Suggest deferring the header write until the first `round_succeeded` round (buffering any leading failed rounds in memory, then flushing header + buffered rows), or otherwise establishing the full result schema up front. A regression test for first-round-failure would lock this in.

limen/experiment/experiment_core.py
```diff
@@ -489,13 +490,18 @@ def _run_with_msq(self,
...
+            data_dict: dict[str, Any] = {}
+
             try:
                 with warnings.catch_warnings(record=True) as caught:
                     warnings.simplefilter('always')
                     data_dict = self.prep(self.data, round_params=sfd_params)
                     round_results = self.model(
                         data=data_dict, round_params=sfd_params,
                     )
+            except StrictModeError as exc:
```
zero-bang, line 532: This `StrictModeError` catch exists only in `_run_with_msq`. The legacy `run()` loop (the no-`search_strategy` path) calls `self.prep(...)` / `self.model(...)` around lines 250-253 with no equivalent try/except, so a strict-mode manifest run without a search strategy will propagate `StrictModeError` and abort the whole experiment instead of recording `strict_mode_error` and continuing — contradicting the documented "UEL catches this per-permutation" behavior in Experiment-Manifest.md. Either mirror the catch in that loop or document that strict mode requires the MSQ path.

zero-bang (changes requested):
Reviewed the full diff plus the surrounding pipeline (CCO loop in `prepare_data`, scaler resolution, both UEL run paths, and the persisted-vs-in-memory result schemas). All 8 checks are green and coverage rises to ~93%, so this is not a CI-gating review.

The core CCO mechanism is sound: prepending `cco_indicator_rows + scaler_context_rows` raw rows of the preceding split, slicing leading indicator nulls, scaling, then dropping exactly the surviving context rows reproduces the continuous-stream scaling to 1e-9 (verified by `test_cco_val_matches_continuous`). The `context_rows = window` math lines up with the `rolling_*(window_size=window).shift(1)` transform. The strict-mode/StrictModeError plumbing and the static/dynamic `extra_params` split are clean and well tested.

Two concerns below, one blocking. Both relate to failure-path bookkeeping in the experiment loop rather than the CCO math itself.

**Engineering note (intended, but broad):** CCO is applied whenever `n_raw_cco > 0`, which includes the indicator warm-up rows even for stateless scalers (`scaler_context_rows == 0` but `cco_indicator_rows > 0`). This means val/test now recover the warm-up rows that previous runs dropped, so val/test row counts — and therefore metrics — shift for *every* existing ML manifest with indicators, not just rolling-scaler ones. That is the correct behavior for inference parity and is tested, but it is a wider behavioral change than the PR title suggests; the Scalers.md line "Stateless scalers ... receive no CCO block" is slightly misleading since they still receive an indicator-warm-up block.

**Forward-looking question:** when a split's preceding context is shorter than the scaler window (`tail(n_raw_cco)` is capped at `len(train)`, e.g. `window=1000` default on a short train split), the scaler cannot be fully warmed and silently falls back to the train-fitted median/IQR via `min_samples` — no error, but val/test no longer match a truly continuous stream. How is that under-warmed case meant to be detected, given strict mode only catches nulls, not warm-up shortfalls?

zero-bang (approved):

Issue #625, opened 2026-06-17
ci: audit tracker for test authority and required gates


mikkokotila:
# CI, test authority, and required gates audit tracker

Source: Limen audit thread `019ecb3f-f4af-7e00-84dd-20bd3e02937e`, final-report extraction from `/private/tmp/limen_audit_final_bullets.txt`.

Scope: test authority and runtime_env gate findings only: pytest/custom runner authority, required checks, matrix coverage, hermeticity, warnings/type/format/package/docs/security gates, and runtime budget scope.

Rule: duplicate audit repeats are collapsed; each distinct finding in this cluster is represented as one checkbox. This issue intentionally does not start implementation work.

### Standard pytest
- [x] **C001 [P0]** Standard `pytest -q` is red while the custom runner is green; contributor-facing test authority is inconsistent.  
  Evidence: Audit lines 76,267,273,275,279,301,695.

### Masked TabPFN test
- [x] **C002 [P0]** `tests/test_tabpfn.py::test_tabpfn` fails because TabPFN resolves to `None`, and the custom runner allowlists/masks it. Split to #628 for real model-download validation.  
  Evidence: Audit lines 76,255,267,279,498.

### Runner authority
- [x] **C003 [P0]** The project has a curated `tests/run.py`, but no public pytest config/contract establishes whether `pytest` or the custom runner is authoritative.  
  Evidence: Audit lines 781,784 plus prior standard pytest findings.

### Runner drift
- [x] **C004 [P1]** Audit found pytest tests outside the custom runner / runner registration drift, so CI can omit test surfaces.  
  Evidence: Audit lines 781,784.

### Warnings gate
- [x] **C005 [P1]** Warnings-as-errors fails; strict warnings are not a required quality gate.  
  Evidence: Audit lines 276,339,415.

### Coverage gate
- [ ] **C006 [P1]** Coverage exists and passes, but branch rules do not require the coverage check.  
  Evidence: Audit lines 184,303,550,754.

### Runtime gate
- [ ] **C007 [P1]** Runtime budget exists, but branch rules do not require runtime check status.  
  Evidence: Audit lines 184,303,550,754.

### Required checks narrow
- [ ] **C008 [P1]** Required checks are only Ruff, Tests, and CodeQL; docs/package/audit/type/format/coverage/runtime/matrix gates are optional.  
  Evidence: Audit lines 184,303,415,550,754,627.

### Python matrix
[...]

Issue #628, opened 2026-06-17
test: split TabPFN model-download proof from gate tracker


mikkokotila:
Source: #625 C002.

## Scope
Validate the real TabPFN model-download path outside the #625 gate-authority branch. #625 may cover lightweight import/configuration behavior only.

## Reason
A proper TabPFN proof requires downloading and initializing the model artifact. That is too heavy and network/cache-sensitive for the CI/test-authority cleanup in #625.

## Repro
1. Install Limen with the `tabpfn` extra in an environment that permits model download.
2. Run `pytest tests/test_tabpfn.py::test_tabpfn -q`.
3. Observe whether TabPFN resolves to a live model implementation instead of `None`.

## Acceptance
- Define the supported TabPFN model-download test environment.
- Add or document the required cache/network contract.
- Provide a reproducible command for validating the real TabPFN path.
- Link the validation evidence back to #625 or its successor tracker.

PR #629, opened 2026-06-17
fix: make pytest the canonical test runner

### LINKED ISSUE #120: feat: move confidence gate into metrics libraries

mikkokotila:
Refs #625. Related: #628.

## What does this PR change?

- Replaces the hand-curated `tests.run` suite list with pytest collection plus runtime-profile emission.
- Adds the public pytest contract in `pyproject.toml`, including warning policy and TabPFN marker registration.
- Makes the heavyweight TabPFN model-download integration test opt-in via `LIMEN_RUN_TABPFN_MODEL_TEST=1`; real model validation is split to #628.
- Bumps Limen to `4.0.1` and updates changelog/template version surfaces.

## Checklist

- [ ] I added and/or updated docstrings as per [Writing Docstrings](https://github.com/Vaquum/Limen/blob/main/docs/Developer/Writing-Docstrings.md) (for any changed public functions/classes)
- [ ] If this PR delivers a PRD’s **last** remaining slice, I also linked the parent PRD to close with it (e.g., “Closes #120”), so the PRD closes on merge instead of by hand

Docstrings are not applicable: no public runtime API was added or changed. This PR intentionally uses `Refs #625`, not `Fixes #625`, because the tracker still includes separate CI policy surfaces.


pyproject.toml
```diff
@@ -122,6 +122,21 @@ relative_files = true
 [tool.coverage.report]
 omit = ["tests/*"]
 
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+addopts = ["-ra"]
+filterwarnings = [
+  "error",
```
zero-bang, line 129: `filterwarnings = ["error", ...]` is a strong, worthwhile gate, but the four sklearn ignores below are pinned to internal module paths (`sklearn.linear_model._logistic`, `sklearn.utils.validation`) and exact message regexes. Those are private modules and unstable warning strings — a routine sklearn upgrade can rename the module or reword the message, at which point the ignore silently stops matching and the warning is promoted to a hard error that fails the whole `Tests` job, with a traceback pointing at sklearn rather than at the upgrade. Non-blocking, but worth either loosening the module qualifier or adding a comment noting these are version-sensitive so the next person knows to re-check them on dependency bumps.

tests/run.py
```diff
@@ -1,2130 +1,101 @@
...
+        suite_finished_at = utc_now_iso()
+        total_duration_seconds = time.perf_counter() - self.suite_started_clock
+        return build_runtime_profile(
+            test_records=self.test_records,
+            suite_started_at=self.suite_started_at,
+            suite_finished_at=suite_finished_at,
+            total_duration_seconds=total_duration_seconds,
+        )
+
+    def _record(self, report: pytest.TestReport) -> None:
```
zero-bang, line 66: `_record` is called unconditionally on `report.when == 'call'` and again on `report.when == 'teardown'` when teardown fails. A test whose `call` phase passes but whose fixture teardown raises will therefore append two records for the same `nodeid` (one `passed`, one `failed`), inflating `test_count`/`passed_count` in the emitted profile and double-counting it for the runtime-budget gate. Rare, but consider skipping the teardown record when a `call` record for the same `nodeid` already exists.

tests/run.py
```diff
@@ -2148,21 +119,54 @@ def _runtime_slowest_tests_limit() -> int:
...
-        logger=logger,
-        profile_output_path=_runtime_profile_output_path(),
-        slowest_tests_limit=_runtime_slowest_tests_limit(),
+def _pytest_args(argv: list[str]) -> list[str]:
+    if argv:
+        return argv
+    return ['tests']
+
+
+def _write_profile(plugin: RuntimeProfilePlugin) -> None:
```
zero-bang: `_write_profile` runs inside `main()`'s `finally` and calls `_runtime_slowest_tests_limit()`, which does `int(configured_limit)` and can `raise ValueError`. If `LIMEN_RUNTIME_SLOWEST_LIMIT` is set to a non-integer (or `<= 0`), this exception fires in the `finally` block *after* the suite has already run, masking pytest's real return code and crashing the runner. Consider validating the env var before `pytest.main` (so a bad value fails fast and clearly) or guarding the slowest-tests logging so it can't take down profile emission.

zero-bang (commented):
## Summary

Clean, well-scoped change. It deletes ~2100 lines of hand-curated suite registration and replaces it with pytest collection plus a `RuntimeProfilePlugin` that emits the same `build_runtime_profile` shape the CI runtime-budget gate already consumes — so the gate keeps working unchanged. `test_run_registration.py` is correctly repurposed from "every test is in the list" to "`run.py` delegates to pytest and emits a profile," and the version bump (pyproject + 5 templates + Experiment-Manifest.md) is consistent with no residual `4.0.0` references. The `pytest.warns(...)` additions in `test_yaml_store.py` are the right consequence of the new `filterwarnings = error` policy. I verified the two new registration tests pass locally and that the full suite collects 1162 tests (the only collection error in my sandbox is a missing `libgomp.so.1` for LightGBM — an environment gap, not a PR defect).

## Open-threads ledger
- No prior reviewer threads.
- **Checks:** Ruff + Package Install (3.10–3.13) pass. `PR Checks Tests`, `Docs Site`, `Packaging`, `CodeQL`, and `Workers Builds` are **pending**. The `Tests` job is the real validator of the new global `filterwarnings = error` policy, which I cannot reproduce locally. I am **deferring** final approval to that check going green rather than approving over a pending gate whose result is the whole point of the PR.

Not requesting changes — the three inline notes below are non-blocking. Engineering note (1.17): the warnings-as-error switch is a genuine quality gate but it couples CI health to dependency warning text; see the pyproject comment. Forward-looking (1.18): the next predictable failure is a routine sklearn/transitive-dep upgrade emitting a *new* warning (or changing an existing message), which will turn the entire `Tests` job red with an error far from the PR that bumped the dependency.

tests/run.py
```diff
@@ -1,2130 +1,101 @@
...
+
+    def pytest_runtest_logreport(self, report: pytest.TestReport) -> None:
+        if report.when == 'call':
+            self._record(report)
+            return
+
+        if report.when == 'setup' and (report.failed or report.skipped):
+            self._record(report)
+            return
+
```
bit-mis, line 52: p3: When the call phase records passed or failed or skipped and the teardown phase then fails, both branches call self._record so the same nodeid is appended twice to test_records. build_runtime_profile then inflates test_count and double-counts that test across passed_count and failed_count for what is logically one test invocation. The same double-recording occurs when setup fails and teardown also fails.

bit-mis (changes requested):

bit-mis (approved):

pyproject.toml
mikkokotila: Addressed in 361e29c by adding a maintenance note above the warning allowlist: these dependency warning filters are version-sensitive and must be re-checked on dependency bumps.

tests/run.py
mikkokotila: Addressed in 361e29c by aggregating runtime profile records by nodeid. A teardown failure now upgrades the existing record to failed instead of appending a second record; regression coverage checks call-plus-teardown failure is counted once.

tests/run.py
mikkokotila: Addressed in 361e29c by aggregating runtime profile records by nodeid. The same test invocation now produces one profile record when call/setup and teardown both report, and regression coverage checks both call-plus-teardown and setup-plus-teardown failure cases.

tests/run.py
mikkokotila: Addressed in 361e29c by validating LIMEN_RUNTIME_SLOWEST_LIMIT before pytest starts and passing the parsed value into profile writing. Invalid values now fail fast with exit code 2 instead of masking pytest results in finally; regression coverage checks this path.

zero-bang (approved):

bit-mis (approved):

